PR #12, opened 2025-05-30
Add auth to contract

ZenGround0:
Needs tests
Conforms to spec: https://www.notion.so/filecoindev/Synapse-Authentication-Spec-201dc41950c1804488b5cd3515fa2f4f

Will be speculatively merging once tests are passing
@aarshkshah1992 when you get back please give it post merge review

### DIFF
--- a/service_contracts/src/SimplePDPServiceWithPayments.sol
+++ b/service_contracts/src/SimplePDPServiceWithPayments.sol
@@ -16,6 +16,14 @@ import {Payments, IArbiter} from "@fws-payments/Payments.sol";
 /// and adjusts payment rates based on storage size. Also implements arbitration
 /// to reduce payments for faulted epochs.
 contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, UUPSUpgradeable, OwnableUpgradeable {
+    // Authentication operations enum
+    enum Operation {
+        CreateProofSet,
+        AddRoots,
+        ScheduleRemovals,
+        DeleteProofSet
+    }
+
     event FaultRecord(uint256 indexed proofSetId, uint256 periodsFaulted, uint256 deadline);
     event ProofSetRailCreated(uint256 indexed proofSetId, uint256 railId, address payer, address payee);
     event RailRateUpdated(uint256 indexed proofSetId, uint256 railId, uint256 newRate);
@@ -45,6 +53,9 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
     // Commission rate in basis points (100 = 1%)
     uint256 public operatorCommissionBps;
 
+    // Mapping from client address to clientDataSetId
+    mapping(address => uint256) public clientDataSetIDs;
+
     // Storage for proof set payment information
     struct ProofSetInfo {
         uint256 railId; // ID of the payment rail
@@ -60,6 +71,7 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
     struct ProofSetCreateData {
         string metadata;
         address payer;
+        bytes signature; // Authentication signature
     }
 
     // Mappings
@@ -190,17 +202,29 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
      * @dev Called by the PDPVerifier contract when a new proof set is created
      * @param proofSetId The ID of the newly created proof set
      * @param creator The address that created the proof set and will receive payments
-     * @param extraData Encoded data containing metadata and payer information
+     * @param extraData Encoded data containing metadata, payer information, and signature
      */
     function proofSetCreated(uint256 proofSetId, address creator, bytes calldata extraData) external onlyPDPVerifier {
-        // Decode the extra data to get the metadata and payer address
+        // Decode the extra data to get the metadata, payer address, and signature
         require(extraData.length > 0, "Extra data required for proof set creation");
         ProofSetCreateData memory createData = decodeProofSetCreateData(extraData);
 
         // Validate the addresses
         require(createData.payer != address(0), "Payer address cannot be zero");
         require(creator != address(0), "Creator address cannot be zero");
-
+        
+        // Generate a new client dataset ID
+        uint256 clientDataSetId = clientDataSetIDs[createData.payer]++;
+        
+        // Verify the client's signature
+        require(
+            verifyCreateProofSetSignature(
+                createData.payer,
+                clientDataSetId,
+                createData.signature
+            ),
+            "Invalid signature for proof set creation"
+        );
         // Initialize the ProofSetInfo struct
         ProofSetInfo storage info = proofSetInfo[proofSetId];
         info.payer = createData.payer;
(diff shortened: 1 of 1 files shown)


service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -190,17 +202,29 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
         require(extraData.length > 0, "Extra data required for proof set creation");
         ProofSetCreateData memory createData = decodeProofSetCreateData(extraData);
 
         // Validate the addresses
         require(createData.payer != address(0), "Payer address cannot be zero");
         require(creator != address(0), "Creator address cannot be zero");
-
+        
+        // Generate a new client dataset ID
+        uint256 clientDataSetId = clientDataSetIDs[createData.payer]++;
```
aarshkshah1992, line 217: This means that `clientDataSetId ` starts at 0 instead of 1. Which is fine but in Solidity, it's usually a good idea for variables to not have the default 0 values to distinguish between initialised/uninitialised values.

How about we start from `clientDataSetId` from 1 instead ?

```
uint256 clientDataSetId = ++clientDataSetIDs[createData.payer];
```

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -578,6 +671,161 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
+     * @notice Recover the signer address from a signature
+     * @param messageHash The signed message hash
+     * @param signature The signature bytes (v, r, s)
+     * @return The address that signed the message
+     */
+    function recoverSigner(
+        bytes32 messageHash,
+        bytes memory signature
+    ) internal pure returns (address) {
+        require(signature.length == 65, "Invalid signature length");
```
aarshkshah1992, line 805: Where does this value of 65 come from ?

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -578,6 +671,161 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
+        bytes32 messageHash = keccak256(
+            abi.encode(
+                address(this),                          // ServiceContractAddr
+                uint8(Operation.DeleteProofSet),        // OpEnum
+                proofSetId                              // ProofSetID
+            )
+        );
+        
+        // Recover signer address from the signature
+        address recoveredSigner = recoverSigner(messageHash, signature);
```
aarshkshah1992, line 789: @ZenGround0 We should add a sanity check that `recoveredSigner  != 0 address` even though we already check for `payer !=0` above

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -578,6 +671,161 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
+     * @param signature The signature bytes (v, r, s)
+     * @return True if the signature is valid, false otherwise
+     */
+    function verifyCreateProofSetSignature(
+        address payer,
+        uint256 clientDataSetId,
+        bytes memory signature
+    ) internal view returns (bool) {
+        // Prepare the message hash that was signed
+        bytes32 messageHash = keccak256(
```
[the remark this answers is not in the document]
aarshkshah1992: We're also not signing the Verifier address. Let's discuss this in the sync today.

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -246,16 +270,47 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
+    /**
+     * @notice Handles proof set deletion and terminates the payment rail
+     * @dev Called by the PDPVerifier contract when a proof set is deleted
+     * @param proofSetId The ID of the proof set being deleted
+     * @param deletedLeafCount The number of leaves deleted
+     * @param extraData Signature for authentication
+     */
+    function proofSetDeleted(
+        uint256 proofSetId,
+        uint256 deletedLeafCount,
```
aarshkshah1992, line 282: nit: comment this param as it's unused and creates a warning.

service_contracts/src/SimplePDPServiceWithPayments.sol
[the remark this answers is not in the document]
aarshkshah1992: We can add this `recoverSigner   != 0 address` check to the `recoverSigner` method itself.

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -578,6 +671,161 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
+        uint256 firstAdded,
+        bytes memory signature
+    ) internal view returns (bool) {
+        // Create a dynamic bytes array to build our message
+        bytes memory message = abi.encode(
+            address(this),                      // ServiceContractAddr
+            uint8(Operation.AddRoots),          // OpEnum
+            proofSetId,                         // ProofSetID
+            firstAdded,                         // FirstAdded
+            rootDataArray                       // Array of rootData
```
aarshkshah1992, line 723: @ZenGround0 Why does a client need to sign the raw size of the data as well if they're already signing the `root cid` ? I don't think they need to sign the entire `rootDataArray ` here.

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -578,6 +671,161 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
+        uint256[] memory rootIds,
+        bytes memory signature
+    ) internal view returns (bool) {
+        // Prepare the message hash that was signed
+        bytes32 messageHash = keccak256(
+            abi.encode(
+                address(this),                          // ServiceContractAddr
+                uint8(Operation.ScheduleRemovals),      // OpEnum
+                proofSetId,                             // ProofSetID
+                rootIds                                 // Array of rootIds
```
aarshkshah1992, line 756: @ZenGround0 Where did we land on how client can lookup rootIds for a given `CommP` ? Do we need add a mapping/event for that in this contract so it's easy for clients to look it up ?

service_contracts/src/SimplePDPServiceWithPayments.sol
aarshkshah1992: Just signing []root Cids should be enough.

service_contracts/src/SimplePDPServiceWithPayments.sol
ZenGround0: I don't see the benefit given we're never signing uninitialized but I also don't mind making the change.

service_contracts/src/SimplePDPServiceWithPayments.sol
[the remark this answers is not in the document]
ZenGround0: Great catch we'll want to sign payee

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -16,6 +16,14 @@ import {Payments, IArbiter} from "@fws-payments/Payments.sol";
 /// and adjusts payment rates based on storage size. Also implements arbitration
 /// to reduce payments for faulted epochs.
 contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, UUPSUpgradeable, OwnableUpgradeable {
+    // Authentication operations enum
+    enum Operation {
```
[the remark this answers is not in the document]
ZenGround0: A single verifier address should be enforced within this service contract so that should not matter.

I think you are right we do want a dataset reference to avoid replays, we should include the clientDataSetID rather than SP address to avoid duplications across the same SP.

PR #16, opened 2025-06-02
Service contract only deployment

ZenGround0:


ZenGround0:
Testing out deployment with this, will merge once its working

PR #18, opened 2025-06-03
All providers getter

ZenGround0:
@rvagg mentioned he needs something that doesn't clobber ethcall apis

### DIFF
--- a/service_contracts/src/SimplePDPServiceWithPayments.sol
+++ b/service_contracts/src/SimplePDPServiceWithPayments.sol
@@ -232,10 +232,36 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
     
     // Getters
     function getAllApprovedProviders() external view returns (ApprovedProviderInfo[] memory) {
-        ApprovedProviderInfo[] memory providers = new ApprovedProviderInfo[](nextServiceProviderId - 1);
-        for (uint256 i = 1; i < nextServiceProviderId; i++) {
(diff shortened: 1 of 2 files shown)


rvagg (approved):
I assume there's no easy way to grow `providers` as we append to it? or we don't maintain a "non-zero providers" count, just to avoid returning a huge list of zeros?

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -229,7 +229,15 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
     function getChallengesPerProof() public pure returns (uint64) {
         return 5;
     }
-
+    
+    // Getters
+    function getAllApprovedProviders() external view returns (ApprovedProviderInfo[] memory) {
+        ApprovedProviderInfo[] memory providers = new ApprovedProviderInfo[](nextServiceProviderId - 1);
+        for (uint256 i = 1; i < nextServiceProviderId; i++) {
+            providers[i - 1] = approvedProviders[i];
```
aarshkshah1992, line 237: @ZenGround0 

This dosent account for the fact that an approved provider can be deleted but `nextServiceProviderId`  always increases.




[9:06](https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1748927219734799?thread_ts=1748924509.699509&cid=C07CGTXHHT4)
But nextServiceProviderId  always increases

rvagg (approved):

PR #19, opened 2025-06-03
Add `withCDN:bool` to `ProofSetInfo`

juliangruber:
This allows consumers like FilCDN to look up whether CDN service has been booked for a proof set.

Next: Also expose in events.

### DIFF
--- a/service_contracts/src/SimplePDPServiceWithPayments.sol
+++ b/service_contracts/src/SimplePDPServiceWithPayments.sol
@@ -261,6 +261,7 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
                 createData.payer,
                 clientDataSetId,
                 creator,
+                createData.withCDN,
                 createData.signature
             ),
             "Invalid signature for proof set creation"
@@ -734,6 +735,7 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
         address payer,
         uint256 clientDataSetId,
         address payee,
+        bool withCDN,
         bytes memory signature
     ) internal view returns (bool) {
         // Prepare the message hash that was signed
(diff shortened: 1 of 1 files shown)


rvagg:
As this is being reviewed, could you also consider how easily accessible this might be with an eth_call - can I get this boolean, what would I need to show up and present to be able to get this boolean from a function call, and what else would I get at the same time? If the data is in the contract state, then having a function an eth_call get fish it out of would be awesome.

juliangruber:
I also added a function :)

```
contract.getProofSetWithCDN(uint256 proofSetId) returns (bool)
```

juliangruber:
PR adding the event: https://github.com/FilOzone/filecoin-services/pull/21

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -270,6 +272,7 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
         info.metadata = createData.metadata;
         info.commissionBps = operatorCommissionBps; // Use the contract's default commission rate
         info.clientDataSetId = clientDataSetId;
+        info.withCDN = createData.withCDN;
```
aarshkshah1992, line 275: @juliangruber @ZenGround0  We should make this field a part of signature verification as well no ?

aarshkshah1992:
lgtm but will wait for @ZenGround0 to review as well.

ZenGround0 (approved):

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -66,12 +66,14 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
         uint256 clientDataSetId; // ClientDataSetID
         mapping(uint256 => string) rootIdToMetadata; // Mapping from root ID to its metadata
+        bool withCDN; // Whether the proof set is using CDN
     }
 
     // Decode structure for proof set creation extra data
     struct ProofSetCreateData {
         string metadata;
         address payer;
+        bool withCDN;
```
jennijuju, line 76: Should this be separate or a part of the metadata? @aarshkshah1992

juliangruber: the metadata string isn't being parsed by the smart contract, I think it's unstructured data (whatever the creator wants to put there). The CDN flag needs to be parsed by the contract.

PR #24, opened 2025-06-03
feat: return client provider info

aarshkshah1992:
Addresses 

```
Get list of providers a client has previously dealt with.

Get list of proof sets for a client with a specific provider (this could be bundled with the above query)
```

For https://github.com/FilOzone/synapse-sdk/issues/56


service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -94,6 +94,11 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
     // Track when proving was first activated for each proof set
     mapping(uint256 => uint256) public provingActivationEpoch;
 
+    // Client-Provider relationship tracking
+    mapping(address => address[]) private clientProviders;
```
ZenGround0, line 98: I don't love adding extra state for ethcalls lets see if we can avoid it.

ZenGround0: Ok the way I think we should do this is add a single mapping from clientDataSetID to proofSetID.  Then in ethcall for a client we iterate all dataset ids get all proofset ids retrieve all proofset infos and build these maps internal to the call.

ZenGround0: We can even return the entire mapping so that the sdk can cache it and avoid more than one call per browser instantiation.

ZenGround0 (approved):
@rvagg FYI that we are including a single accessor for all of a client's proofsets so that SDK can get access to all the info it is going to need in one call.  We explored exposing only the methods you requested and they had issues with either 1) cluttering the state or 2) being impractically complicated to implement in solidity given our very limited set of in memory datastructures (i.e. no lists or hashtables)

rvagg:
👍 definitely don't want to clutter state for this; I'll experiment with this and report back. Thanks

PR #27, opened 2025-06-04
feat(test): add auth sig fixture generator & tester

rvagg:
Generates output like this:

```json
  {
    "signer": "0x2e988A386a799F506693793c6A5AF6B54dfAaBfB",
    "createProofSet": {
      "signature": "0x7486153ac70abc0b45f07cfa3d966505bce69104d277402f5111a54c36c0266222947952867ac4365ae79263a04639bb2ff9c2a92b12e54b9b910fe2ff89fc7c1b",
      "clientDataSetId": 12345,
      "payee": "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
      "withCDN": true
    },
    "addRoots": {
      "signature": "0xc939c6e9d1c685cf23f966f3b462b9db48b280533e99e44577d397ba43fe917c03b8ab99c2650f203f7479ee3bf9f1bc259505fb30654d7dd920eac6950a77261c",
      "clientDataSetId": 12345,
      "firstAdded": 1,
      "rootDigests": ["0xfc7e928296e516faade986b28f92d44a4f24b935485223376a799027bc18f833", "0xa9eb89e9825d609ab500be99bf0770bd4e01eeaba92b8dad23c08f1f59bfe10f"],
      "rootSizes": [1024, 2048]
    },
    "scheduleRemovals": {
      "signature": "0xdd0c885ebd532034fa1b442193ea6f6b6480d8fbcf1d29dc5eae0cc2eeea468850ea022d2693009231aff0a1dfd59f69adf6389a771f6e6d88e0e4a8d9b1f4e31b",
      "clientDataSetId": 12345,
      "rootIds": [1, 3, 5]
    },
    "deleteProofSet": {
      "signature": "0x614385d6a5fe4c3bc4693adf25fa4e81ec64af49c8ea40e058cb90f91c026d305083801f25a84b1dcfa9525034aaeaeadca89a5f5fb01b2257d59d2c2b99f4701b",
      "clientDataSetId": 12345
    }
  }
```

Which it can also read and verify.

1. Generate reference signatures: `forge test --match-test testGenerateFixtures -vv`
2. Copy JSON to ./test/external_signatures.json
3. Verify round-trip works: `forge test --match-test testExternalSignatures -vv`

I should be able to use this output in the SDK to do matching, or put my own output into test/external_signatures.json and verify.


rvagg:
Fixed the contract in here too in the latest commit. Things I learned along the way:

1: For the typehash to identify the type being signed, where that type is a compound that is made up of multiple types, such as in `AddRoots` which includes `RootData` which includes `Cid` (i.e. the final type to encode has 3 types in total), the types being encoded should be sorted before being encoded.

> If the struct type references other struct types (and these in turn reference even more struct types), then the set of referenced struct types is collected, sorted by name and appended to the encoding.
[ref](https://eips.ethereum.org/EIPS/eip-712#definition-of-encodetype)

So in our case:

`"AddRoots(uint256 clientDataSetId,uint256 firstAdded,RootData[] rootData)RootData(Cid root,uint256 rawSize)Cid(bytes data)"`

should actually be:

`"AddRoots(uint256 clientDataSetId,uint256 firstAdded,RootData[] rootData)Cid(bytes data)RootData(Cid root,uint256 rawSize)"`

2: When we encode those types, where the array length is dynamic in length, we can’t just use `abi.encode()` which will length-prefix it, we just need them to be concatenated together so we must use `abi.encodePacked()` to do it and avoid the length prefix.

aarshkshah1992:
@rvagg If you are happy with my commit  to encode packed `rootIds` as well-> I can update the test and fixtures accordingly.

PR #28, opened 2025-06-04
Move to EIP 712

### LINKED ISSUE #26: MetaMask signature compatibility for SimplePDPServiceWithPayments
The current contract expects **raw signatures** (signed message hash directly), but MetaMask and other external wallets automatically add the **Ethereum message prefix** when signing, which changes the verification hash.

```js
keccak256("\x19Ethereum Signed Message:\n" + message.length + message)
```

See https://eips.ethereum.org/EIPS/eip-191

## Option 1: Dual Signature Support

This would make it possible to support unprefixed messages _or_ prefixed messages. It has complexity overhead and a slight gas cost increase.

Modify signature verification functions to support both raw and prefixed signatures:

```solidity
// Add this import
import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";

function verifyCreateProofSetSignature(
    address payer,
    uint256 clientDataSetId,
    address payee,
    bool withCDN,
    bytes memory signature
) internal view returns (bool) {
    bytes32 messageHash = keccak256(
        abi.encode(
            address(this),
            uint8(Operation.CreateProofSet),
            clientDataSetId,
            withCDN,
            payee
        )
    );
    
    // Try prefixed signature first (MetaMask/external wallets)
    bytes32 ethSignedMessageHash = ECDSA.toEthSignedMessageHash(messageHash);
    address recoveredSigner = ECDSA.recover(ethSignedMessageHash, signature);
    
    // If prefixed signature fails, try raw signature (SDK with private keys)
    if (recoveredSigner != payer) {
        recoveredSigner = ECDSA.recover(messageHash, signature);
    }
    
    return recoveredSigner == payer;
}
```

Apply the same pattern to all signature verification functions:
- `verifyAddRootsSignature`
- `verifyScheduleRemovalsSignature` 
- `verifyDeleteProofSetSignature`

## Option 2: Force Prefixed Signatures Only

This is the simplest option, a breaking change (that's fine), gas-cheap, and the prefixed message format can be easily emulated even without MetaMask so this format can be produced either way.

Modify the contract to only accept prefixed signatures:

```solidity
function verifyCreateProofSetSignature(
    address payer,
    uint256 clientDataSetId,
    address payee,
    bool withCDN,
    bytes memory signature
) internal view returns (bool) {
    bytes32 messageHash = keccak256(
        abi.encode(
            address(this),
            uint8(Operation.CreateProofSet),
            clientDataSetId,
            withCDN,
            payee
        )
    );
    
    // Always use prefixed hash
    bytes32 ethSignedMessageHash = ECDSA.toEthSignedMessageHash(messageHash);
    address recoveredSigner = ECDSA.recover(ethSignedMessageHash, signature);
    
    return recoveredSigner == payer;
}
```

---

It would be good to document this well in the contract whichever way we choose to deal with this.

ZenGround0:
Closes #26 

I decided against adding solidity tests since the actual test that matters will be integration with front end / metamask code.

### DIFF
--- a/service_contracts/src/SimplePDPServiceWithPayments.sol
+++ b/service_contracts/src/SimplePDPServiceWithPayments.sol
@@ -7,6 +7,7 @@ import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import "@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol";
 import {Payments, IArbiter} from "@fws-payments/Payments.sol";
 
 /// @title SimplePDPServiceWithPayments
@@ -15,14 +16,7 @@ import {Payments, IArbiter} from "@fws-payments/Payments.sol";
 /// using the Payments contract. It creates payment rails for storage providers
 /// and adjusts payment rates based on storage size. Also implements arbitration
 /// to reduce payments for faulted epochs.
-contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, UUPSUpgradeable, OwnableUpgradeable {
-    // Authentication operations enum
-    enum Operation {
-        CreateProofSet,
-        AddRoots,
-        ScheduleRemovals,
-        DeleteProofSet
-    }
+contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, UUPSUpgradeable, OwnableUpgradeable, EIP712Upgradeable {
 
     event FaultRecord(uint256 indexed proofSetId, uint256 periodsFaulted, uint256 deadline);
     event ProofSetRailCreated(uint256 indexed proofSetId, uint256 railId, address payer, address payee, bool withCDN);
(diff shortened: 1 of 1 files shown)


service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -15,14 +16,7 @@ import {Payments, IArbiter} from "@fws-payments/Payments.sol";
 /// using the Payments contract. It creates payment rails for storage providers
 /// and adjusts payment rates based on storage size. Also implements arbitration
 /// to reduce payments for faulted epochs.
-contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, UUPSUpgradeable, OwnableUpgradeable {
-    // Authentication operations enum
-    enum Operation {
```
ZenGround0, line 20: Note that Operation is redundant with the EIP 712 named type hashes for structs.  Also specifying the verifying contract is redundant with the EIP 712 domain separator.  So we can simplify our custom structs a good amount.

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -7,6 +7,7 @@ import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import "@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol";
```
ZenGround0, line 10: Please note we're using upgradeable so we can initialize the contract address properly

service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -131,6 +125,31 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
     event ProviderRejected(address indexed provider);
     event ProviderRemoved(address indexed provider, uint256 indexed providerId);
 
+    // EIP-712 Type hashes
+    bytes32 private constant CREATE_PROOFSET_TYPEHASH = keccak256(
+        "CreateProofSet(uint256 clientDataSetId,bool withCDN,address payee)"
```
rvagg, line 130: so `address(this)` is implied now? and obviously no need for the `Operation` `uint8`

ZenGround0: Yeah exactly the EIP 712 spec specifies contract address as one of the several values that get hashed as part of the domain separator used in all message digests.  If you click through the  _hashTypedDataV4 you can see where that shows up.

PR #29, opened 2025-06-04
feat: differential pricing policy if client wants CDN retrievals

### LINKED ISSUE #25: Make sure pricing is different if the `withCDN` flag is enabled
We need to this once https://github.com/FilOzone/filecoin-services/pull/19 lands.

aarshkshah1992:
Closes #25  .

### DIFF
--- a/service_contracts/src/SimplePDPServiceWithPayments.sol
+++ b/service_contracts/src/SimplePDPServiceWithPayments.sol
@@ -73,6 +73,14 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
         bytes signature; // Authentication signature
     }
 
+    // Structure for service pricing information
+    struct ServicePricing {
+        uint256 pricePerGiBPerMonthNoCDN;  // Price without CDN (2 USDFC per GiB per month)
+        uint256 pricePerGiBPerMonthWithCDN; // Price with CDN (3 USDFC per GiB per month)
+        address tokenAddress;               // Address of the USDFC token
+        uint256 epochsPerMonth;             // Number of epochs in a month
+    }
+
     // Mappings
     mapping(uint256 => uint256) public provingDeadlines;
     mapping(uint256 => bool) public provenThisPeriod;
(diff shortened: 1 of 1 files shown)


service_contracts/src/SimplePDPServiceWithPayments.sol
```diff
@@ -685,29 +695,27 @@ contract SimplePDPServiceWithPayments is PDPListener, IArbiter, Initializable, U
...
         uint256 ratePerEpoch = numerator / denominator;
 
-        // Ensure minimum rate is 0.00001 USDFC if calculation results in 0 due to rounding,
-        // but only if bytes >= 1 MiB (already checked above).
-        // This prevents charging 0 for sizes slightly above 1 MiB but below the threshold for a rate of 1 unit.
-        if (ratePerEpoch == 0 && totalBytes >= MIB_IN_BYTES) {
+        // Ensure minimum rate is 0.00001 USDFC if calculation results in 0 due to rounding.
+        // This prevents charging 0 for very small sizes due to integer division.
+        if (ratePerEpoch == 0 && totalBytes > 0) {
             uint256 minRate = (1 * 10 ** uint256(tokenDecimals)) / 100000;
```
rvagg, line 719: `1 *`? is that doing any work here? should this be doing a `* 100000` as well for symmetry in the calculation to get back to the base rate?

rvagg: oh, nevermind about the `*100000` thing, I see that this is literally doing a 0.0001 base rate; that's fine, but what about the `1 *`? is that just to make it clear that it's not 0.000X.?

feel free to ignore my comment if this is all intended

rvagg:
fine I think, except for the use of magic numbers repeated, also assuming GiB is the right amount for these numbers?

rjan90:
> also assuming GiB is the right amount for these numbers?

According to https://www.notion.so/filecoindev/Specify-the-feature-sets-to-build-for-the-hackathon-M1-1fbdc41950c180ca877fe1e772abe562?source=copy_link

Pricing should be:

> - Basic product: 2 USDFC/tib/month. Data will be retrieved directly from the SPs.
- Add-on: 3 USDFC for everything offered in basic product, with a CDN *add-on* and data will be retrieved from the CDN. ‣

rvagg:
ok, so TiB was right originally, better stick to that then!

rjan90:
> The rate doesnt seem to be implementing what was suggested [here](https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1748392686006789)

@jennijuju I'm wondering if this is a hard requirement for M1. While I recognize that it could provide valuable insights into the fee-switch functionality and its workflows, I'm questioning the priority given our current focus on resolving critical issues to achieve E2E functionality.

It would be helpful to settle some of the moving pieces first so we can narrow our focus and make more targeted progress. We can upgrade the contract after M1 with this fee-functionality.

rjan90:
> Currently blocked by an outstanding question of having the fee-switch implemented. Wondering if this is a hard requirement for M1?

From the sync today; is that @jennijuju might submit a PR, but it is not blocking this PR. @aarshkshah1992 once the conflict has been resolved we can merge this.

PR #34, opened 2025-06-05
Debug/failing create proofset

ZenGround0:
FYI this was the code for debugging the contract today.  Other than the fix in `decodeProofSetCreateData ` I don't think anything here needs to stay but maybe useful for continued debugging.


ZenGround0:
Ah looks like this is no longer needed

PR #35, opened 2025-06-05
proofset creation: fix decode issue for extradata

aarshkshah1992:
pulls it out from https://github.com/FilOzone/filecoin-services/pull/34/files


rvagg:
Oh, tests need fixing

PR #37, opened 2025-06-05
feat: different pricing for CDN vs no CDN

aarshkshah1992:
Implements different pricing for CDN vs no CDN


aarshkshah1992:
Subsumes https://github.com/FilOzone/filecoin-services/pull/29.

rjan90 (approved):

PR #42, opened 2025-06-07
feat: add getProofSet to retrieve proofset details safely

rvagg:
Getting a struct from a map is apparently not stable due to memory packing concerns. I'm having a lot of trouble getting it directly from proofSetInfo, so this getter should make it safe and stable.


rvagg:
Locally this fails for me but afaik I didn't touch anything related to it:

```
[FAIL: next call did not revert as expected] testDirectImplementationCallFails() (gas: 104763)
```

rvagg:
https://github.com/ethers-io/ethers.js/discussions/3484#discussioncomment-4030219 interesting reference about this 🤷

aarshkshah1992 (approved):

PR #43, opened 2025-06-09
feat: quick method for adding SPs to the contract

aarshkshah1992:
A one step "register + approve" that owners can approve.


jennijuju:
@rjan90 can we make sure to have a todo / issue to remove it post m1 (shouldn’t be there on mainnet afaict

PR #46, opened 2025-06-10
docs: add readme

jennijuju:


rjan90:
> ok, but there is a Makefile in here to that runs tests and can clean, so maybe the section that tells you to use foundry could be updated or the makefile removed, or ..

I created a ticket for us to follow up on this here: https://github.com/FilOzone/filecoin-services/issues/49, so that this question does not block merging this

Issue #67, opened 2025-07-02
Refactor MinerInfo after PDP → IPNI support is added in curio


jennijuju:
**Description**:
Once PDP → IPNI support is implemented, we should consider the following changes to reduce on-chain/off-chain discrepancies and streamline SP key management:

**Multiaddr Management:**

SPs server url is published as mutladdr to IPNI and curio/ipni ensure it is updated whenever the SP changes their server URL or endpoint.
Remove the on-chain `uploadUrl` and `downloadUrl` fields from the service contract to minimize the risk of data drift and inconsistencies between on-chain records and IPNI-published information.

**Peer ID Simplification:**

Add a peerId field that is derived from the SP `control walle`t public key. This reduces the number of keys SP must manage, and ensures a tighter link between service identity, content routing identifier and sp ops wallet.

**Open Questions:**

- Are there any edge cases where on-chain URLs are still needed? cc @juliangruber, @bajtos  ?
- Is there consensus on which public key to use for peerId derivation (i.e., always control wallet)? cc @LexLuthr

jennijuju:
one of the good reason to keep a server url is to ensure `/upload` can happen seamlessly without offchain/external dependencies

after chatting with @lexluthr, we are aligned to keep one server url here and curio (or other sp software) is in charge of making the onchain &offchain url consistent for SPs

Issue #62, opened 2025-06-23
Storage Provider Information Modification Flow


rjan90:
## Summary
SPs currently cannot modify their provider information (PDP URL, piece retrieval URL) after being approved. Once a provider is approved, the only way to change their information is to be removed by an admin and re-register from scratch.

## Current Behavior
The current provider management flow only supports:

1. Initial Registration: SP calls [registerServiceProvider(pdpUrl, pieceRetrievalUrl)](https://github.com/FilOzone/filecoin-services/blob/1b99d8b39bf7f35608884b0bd7bd768810ee2372/service_contracts/src/PandoraService.sol#L1078)
1. Admin Review: Admin calls [approveServiceProvider(provider)](https://github.com/FilOzone/filecoin-services/blob/1b99d8b39bf7f35608884b0bd7bd768810ee2372/service_contracts/src/PandoraService.sol#L1099) or [rejectServiceProvider(provider)](https://github.com/FilOzone/filecoin-services/blob/1b99d8b39bf7f35608884b0bd7bd768810ee2372/service_contracts/src/PandoraService.sol#L1132-L1133)
1. Admin Removal: Admin calls r[emoveServiceProvider(providerId)](https://github.com/FilOzone/filecoin-services/blob/1b99d8b39bf7f35608884b0bd7bd768810ee2372/service_contracts/src/PandoraService.sol#L1152) to revoke approval

## Requested Feature
Add a modification flow that allows approved providers to update their information:
1. **SP Modification Request**: modifyServiceRequest(newPdpUrl, newPieceRetrievalUrl)
- Only callable by already approved providers
- Stores the new information in a pending modification state
- Emits ProviderModificationRequested event
2. Admin Review: approveModifyServiceRequest(provider)
- Only callable by contract owner/admin
- Overrides the existing provider info with the new pending information
- Emits ProviderModificationApproved event

## Use Cases
- URL Changes: Providers need to update their service endpoints due to infrastructure changes
- Service Improvements: Providers want to migrate to new domains or improved infrastructure
- Operational Updates: Providers need to update URLs for maintenance or scaling

## Implementation Considerations
1. State Management:
- Add PendingModificationInfo struct similar to PendingProviderInfo
- Add mapping for pending modifications: mapping(address => PendingModificationInfo) public pendingModifications

2. Access Control:
- modifyServiceRequest() should only be callable by approved providers
- approveModifyServiceRequest() should only be callable by admin/owner

[...]

PR #77, opened 2025-07-07
refactor!: update and cleanup provider info/management  

### LINKED ISSUE #47: Remove quick SP registration method post-M1 testing
PR #43 introduced a quick method for adding Storage Providers (SPs) to the contract that combines "register + approve" functionality for owners to approve in a single step. This convenience method should not be present on mainnet.

## Done Criteria
- Revert #43 after M1 testing has been completed
   - Estimated date M1 testing has been completed would be 2025-06-XX.
- No breaking changes for existing properly registered SPs

jennijuju:
1. Removed addServiceProvider function;  resolves https://github.com/FilOzone/filecoin-services/pull/47
  2. Updated provider structures (providerServiceUrl + peerId) (partial of #67)        - https://github.com/FilOzone/filecoin-services/issues/62 seems dependent on this
 
~  3. allow batch approval, removal and rejection to be gas/operation efficient. note: partial success is allowed ~

this may need subgraph updates

### DIFF
--- a/service_contracts/src/PandoraService.sol
+++ b/service_contracts/src/PandoraService.sol
@@ -1213,37 +1213,6 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         return providerToId[provider];
     }
     
-    /**
-     * @notice Add a service provider directly without registration process
-     * @dev Only owner can add providers directly. This bypasses the register+approve flow.
-     * @param provider The address of the provider to add
-     * @param pdpUrl The URL for PDP services
-     * @param pieceRetrievalUrl The URL for piece retrieval services
-     */
-    function addServiceProvider(address provider, string calldata pdpUrl, string calldata pieceRetrievalUrl) external onlyOwner {
-        require(provider != address(0), "Provider address cannot be zero");
-        require(!approvedProvidersMap[provider], "Provider already approved");
-        
-        // Assign ID and store provider info
-        uint256 providerId = nextServiceProviderId++;
-        approvedProviders[providerId] = ApprovedProviderInfo({
-            owner: provider,
-            pdpUrl: pdpUrl,
-            pieceRetrievalUrl: pieceRetrievalUrl,
-            registeredAt: block.number,
-            approvedAt: block.number
-        });
-        
-        approvedProvidersMap[provider] = true;
-        providerToId[provider] = providerId;
-        
-        // Clear any pending registration if it exists
-        if (pendingProviders[provider].registeredAt > 0) {
-            delete pendingProviders[provider];
-        }
-        
-        emit ProviderApproved(provider, providerId);
-    }
 
     function getClientProofSets(address client) public view returns (ProofSetInfo[] memory) {
         uint256[] memory proofSetIds = clientProofSets[client];
--- a/service_contracts/src/PandoraService.sol
+++ b/service_contracts/src/PandoraService.sol
@@ -116,14 +116,14 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         
     struct ApprovedProviderInfo {
         address owner;
-        string serviceUrl; // HTTP server URL for provider services; TODO: Standard API endpoints:{providerServiceUrl}/api/upload / {providerServiceUrl}/api/info 
+        string serviceURL; // HTTP server URL for provider services; TODO: Standard API endpoints:{serviceURL}/api/upload / {serviceURL}/api/info 
         bytes peerId; // libp2p peer ID (optional - empty bytes if not provided)
         uint256 registeredAt; 
         uint256 approvedAt;
     }
     
     struct PendingProviderInfo {
-        string providerServiceUrl; // HTTP server URL for provider services; TODO: Standard API endpoints:{providerServiceUrl}/api/upload / {providerServiceUrl}/api/info 
+        string serviceURL; // HTTP server URL for provider services; TODO: Standard API endpoints:{serviceURL}/api/upload / {serviceURL}/api/info 
         bytes peerId; //libp2p peer ID (optional - empty bytes if not provided)
         uint256 registeredAt;
     }
@@ -141,7 +141,7 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
     uint256 public challengeWindowSize;
     
     // Events for SP registry
-    event ProviderRegistered(address indexed provider, string providerServiceUrl, bytes peerId);
+    event ProviderRegistered(address indexed provider, string serviceURL, bytes peerId);
     event ProviderApproved(address indexed provider, uint256 indexed providerId);
     event ProviderRejected(address indexed provider);
     event ProviderRemoved(address indexed provider, uint256 indexed providerId);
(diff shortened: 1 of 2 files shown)


service_contracts/src/PandoraService.sol
```diff
@@ -116,16 +116,17 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
...
     }
     
     struct PendingProviderInfo {
-        string pdpUrl;
-        string pieceRetrievalUrl;
-        uint256 registeredAt; 
+        string providerServiceUrl; // HTTP server URL for provider services; TODO: Standard API endpoints:{providerServiceUrl}/api/upload / {providerServiceUrl}/api/info 
+        bytes peerId; //libp2p peer ID (optional - empty bytes if not provided)
+        uint256 registeredAt;
+
```
rvagg, line 129:
```suggestion
```

service_contracts/src/PandoraService.sol
```diff
@@ -116,16 +116,17 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         
     struct ApprovedProviderInfo {
         address owner;
-        string pdpUrl;
-        string pieceRetrievalUrl;
+        string providerServiceUrl; // HTTP server URL for provider services; TODO: Standard API endpoints:{providerServiceUrl}/api/upload / {providerServiceUrl}/api/info 
+        bytes peerId; //libp2p peer ID (optional - empty bytes if not provided)
```
rvagg, line 120:
```suggestion
        bytes peerId; // libp2p peer ID (optional - empty bytes if not provided)
```

jennijuju:
I wanted to clarify the reasoning behind dropping retrieveURL. As I recall (need to find where this was discussed), the main driver was feedback from both Curio (cc @LexLuthr) and storage providers: Curio only supports configuration of a single server URL for SPs, expecting them to handle proxies or internal routing if they want separate endpoints for upload and retrieval. Conversations with other SPs indicated they also prefer service/base URLs to remain static for simplicity and operational clarity.

Given this, my proposal is to standardize the route/path structure rather than add more complexity to the base config onchain:

- {serviceURL}/api/upload for POST (data push)
- {serviceURL}/api/piece for GET (retrieval)
- {serviceURL}/contact, {serviceURL}/info, etc., for service metadata (as standards emerge in the tooling)

I understand your point about retriever diversity and resilience if a content routing provider goes down, and agree it's important. That said, this  approach means:
- Clients know exactly where to find interfaces on any compliant provider, reducing guesswork
- It preserves compatibility with SubgraphRetriever and ChainRetriever, which I agree is very important, And it sidesteps the need for additional moving parts in config and content routing provider dependencies
- If a provider wants to do something more complex, they can still proxy or load-balance behind their static serviceURL. This keeps the system predictable while not limiting operational flexibility on the SP side.

lmk if this sounds reasonable - if we still have strong disagreements, im happy to separate commit 2 to a separate thing

jennijuju:
>  but only a peerId that we have to go fishing around in some third source

@rvagg also just want to confirm i did keep a providerServiceUrl for reasons mentioned above and in [this](https://github.com/FilOzone/filecoin-services/issues/67) instead of including a single peerID

service_contracts/src/PandoraService.sol
```diff
@@ -1071,106 +1072,195 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
...
+            address provider = providers[i];
+            
+            // Skip if no pending registration or already approved
+            if (pendingProviders[provider].registeredAt == 0 || approvedProvidersMap[provider]) {
+                continue;
+            }
+            
+            // Update mappings
+            approvedProvidersMap[provider] = false;
+            providerToId[provider] = 0;
```
silent-cipher, line 1190: Nit: These lines might be redundant, since `approvedProvidersMap[provider]` is already `false` by this point, and `providerToId[provider]` wouldn't exist in either `registered` or `removed` cases

silent-cipher (approved):
Approving from my side!

service_contracts/src/PandoraService.sol
```diff
@@ -116,16 +116,17 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         
     struct ApprovedProviderInfo {
         address owner;
-        string pdpUrl;
-        string pieceRetrievalUrl;
+        string providerServiceUrl; // HTTP server URL for provider services; TODO: Standard API endpoints:{providerServiceUrl}/api/upload / {providerServiceUrl}/api/info
```
juliangruber, line 119:
```suggestion
        string serviceUrl; // HTTP server URL for provider services; TODO: Standard API endpoints:{providerServiceUrl}/api/upload / {providerServiceUrl}/api/info 
```

Redundant naming

service_contracts/src/PandoraService.sol
```diff
@@ -1071,106 +1072,195 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
...
-     * @notice Approve a pending service provider
-     * @dev Only owner can approve providers
-     * @param provider The address of the provider to approve
+     * @notice Approve pending service providers
+     * @dev Only owner can approve providers. Pass single-element array for one provider.
+     * @param providers Array of provider addresses to approve
+     * @return successfulProviders Array of provider addresses that were successfully approved
+     * @return successfulProviderIds Array of provider IDs that were assigned to approved providers
      */
-    function approveServiceProvider(address provider) external onlyOwner {
```
juliangruber, line 1099: Are we really concerned about approving so many providers that gas is an issue? I'm wondering if the complexity here is justified

juliangruber (changes requested):

service_contracts/src/PandoraService.sol
jennijuju: I hear you - the state loading and writing is small here and the additional codes here may be unnecessary, i will remove this change

rvagg (approved):

service_contracts/src/PandoraService.sol
```diff
@@ -116,16 +116,16 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         
     struct ApprovedProviderInfo {
         address owner;
-        string pdpUrl;
-        string pieceRetrievalUrl;
+        string serviceURL; // HTTP server URL for provider services; TODO: Standard API endpoints:{serviceURL}/api/upload / {serviceURL}/api/info
```
bajtos, line 119: How can Synapse SDK and FilCDN derive the URL for Piece retrievals? Is it `{serviceUrl}/piece/{CommP}`?

Can you please include this endpoint in the code comment?

service_contracts/src/PandoraService.sol
```diff
@@ -141,7 +141,7 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
     uint256 public challengeWindowSize;
     
     // Events for SP registry
-    event ProviderRegistered(address indexed provider, string pdpUrl, string pieceRetrievalUrl);
+    event ProviderRegistered(address indexed provider, string serviceURL, bytes peerId);
     event ProviderApproved(address indexed provider, uint256 indexed providerId);
```
bajtos, line 145: While we are making this change, would you @jennijuju mind to modify `ProviderApproved` event to include `string serviceURL, bytes peerId` in the payload?

This would significantly simplify our life, potentially also life of other consumers of these events.

Right now, we are listening for `ProviderApproved` and `ProviderRemoved` events to maintain a list of approved SPs _and_ their service URLs.

Because `ProviderApproved` event does not include any info about the provider, we have to make another RPC API call in the ProviderApproved event handler to fetch that data from the on-chain state.

If the event includes the info, we can get all the data we need from the event payload.

service_contracts/src/PandoraService.sol
rvagg: yes, I think for now the comment is wrong here, retrievals will be `{serviceURL}/piece/{commp}` and pdp interactions will be `{serviceURL}/pdp/...`; we should probably be careful with docs here mentioning `/api/` because that's not a thing atm, it's just /pdp/ and /piece/ that we work with.

rvagg:
```suggestion
        string serviceURL; // HTTP server URL for provider services; TODO: specify standard API endpoints with FRC
```

service_contracts/src/PandoraService.sol
```diff
@@ -116,16 +116,16 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
...
         uint256 registeredAt; 
-        uint256 approvedAt;   
+        uint256 approvedAt;
     }
     
     struct PendingProviderInfo {
-        string pdpUrl;
-        string pieceRetrievalUrl;
-        uint256 registeredAt; 
+        string serviceURL; // HTTP server URL for provider services; TODO: Standard API endpoints:{serviceURL}/api/upload / {serviceURL}/api/info
```
rvagg, line 126:
```suggestion
        string serviceURL; // HTTP server URL for provider services; TODO: specify standard API endpoints with FRC
```

juliangruber (approved):

bajtos:
🎉

Issue #59, opened 2025-06-23
Set Service Creator Fee to 0% Commission


rjan90:
## Description
Update the service creator fee configuration to set the commission fee to 0%.

PR #78, opened 2025-07-08
chore: set basic service commission to 0% (was 5%)

rjan90:
Updates the service creator fee configuration to set the basic service commission to 0% as requested in #59.


rjan90:
Ah, I should have caught that the tests were failing. Pushed a fix for those now

PR #79, opened 2025-07-09
 Implement Key-Value Metadata Structure

### LINKED ISSUE #63: Implement Key-Value Metadata Structure
## Summary

> This initially started out as a discussion, please see the "Conclusion" section of this issue ticket, together with the discussion comments. Keeping the initial summary/background/discussion points for historical reference from where this discussion originated from.

Discussion about the design choice for the metadata field in the extraData structure - whether it should remain unstructured (current implementation) or become structured to enable better tooling support.

## Background

From PR discussion in [FilOzone/filecoin-services#19](https://github.com/FilOzone/filecoin-services/pull/19/files#r2125151537) and [this Slack Thread](https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1749039857407669?thread_ts=1748957215.157949&cid=C07CGTXHHT4) in the #fil-services channel:

- The metadata field was added during HotVault development for users to add string labels/identifiers for payment rails
- Currently designed as unstructured data for "opaque identifiers" (similar to a "note" field in payments)
- Example use case: Adding "deal-for-cat-picture" to be displayed in Explorer UI
- Never fully utilized in current implementation

### Current Design (Unstructured)
- **Pros**: Flexible, no need for upfront standardization, simple implementation
- **Use case**: User-defined labels/notes for deals/proofsets/roots that appear in Explorer
- **Implementation**: Part of extraData, emitted in events for indexer consumption

### Proposed Concern (Need for Structure)
- **Issue**: Unstructured metadata creates friction for tooling development
- **Problem**: Difficult for tools like explorers to parse and display service information consistently
- **Example**: Product descriptions, CDN pricing info not standardized
- **Risk**: Fragmented ecosystem where each service uses different metadata formats

## Conclusion

> This is an update on 2025-06-26 after the discussion that has been going on the thread below.

We have converged on [rvagg's key-value pair proposal ](https://github.com/FilOzone/filecoin-services/issues/63#issuecomment-3002705793) after iterating through different perspectives.

### Technical Implementation:
- Use mapping(uint256 => mapping(string => bytes)) public proofSetMetadata
- Support metadataKeys[] and metadataValues[] arrays
- Emit events for indexer consumption

### Next Steps:
- Implement the key-value approach in both contract and SDK
- Extend same pattern to `AddRoots` functionality

pali101:
## Summary

- Switched proof set metadata to a key-value model (`mapping(uint256 => mapping(string => bytes))`).
- Updated all function and structs to use `metadataKeys[]` and `metadataValues[]` instead of string-based metadata.
- Updated the test suite to cover the new structure.
- Added `testProofSetMetadataStorage`, `testProofSetMetadataStorageMultipleKeys`, and `testProofSetMetadataStorageMultipleProofSets`.

Closes #63

### DIFF
--- a/service_contracts/src/PandoraService.sol
+++ b/service_contracts/src/PandoraService.sol
@@ -22,6 +22,13 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
     event ProofSetRailCreated(uint256 indexed proofSetId, uint256 railId, address payer, address payee, bool withCDN);
     event RailRateUpdated(uint256 indexed proofSetId, uint256 railId, uint256 newRate);
     event RootMetadataAdded(uint256 indexed proofSetId, uint256 rootId, string metadata);
+    event ProofSetCreatedWithMetadata(
+        uint256 indexed proofSetId,
+        address creator,
+        address payer,
+        string[] metadataKeys,
+        bytes[] metadataValues
+    );
 
     // Constants
     uint256 public constant NO_CHALLENGE_SCHEDULED = 0;
@@ -60,6 +67,11 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
     mapping(address => uint256) public clientDataSetIDs;
     // Mapping from proof set ID to root ID to metadata
     mapping(uint256 => mapping(uint256 => string)) public proofSetRootMetadata;
+    // Mapping from proof set ID to key value pair metadata
+    // proofSetId => (key => value)
+    mapping(uint256 => mapping(string => bytes)) public proofSetMetadata;
+    // proofSetId => array of keys
+    mapping(uint256 => string[]) internal proofSetMetadataKeys;
 
     // Storage for proof set payment information
     struct ProofSetInfo {
@@ -67,7 +79,7 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         address payer; // Address paying for storage
         address payee; // SP's beneficiary address
         uint256 commissionBps; // Commission rate for this proof set (dynamic based on whether the client purchases CDN add-on)
-        string metadata; // General metadata for the proof set
+        // string metadata; // General metadata for the proof set
         string[] rootMetadata; // Array of metadata for each root
         uint256 clientDataSetId; // ClientDataSetID
         bool withCDN; // Whether the proof set is registered for CDN add-on
@@ -75,8 +87,9 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
 
     // Decode structure for proof set creation extra data
     struct ProofSetCreateData {
-        string metadata;
         address payer;
+        string[] metadataKeys;
+        bytes[] metadataValues;
         bool withCDN;
         bytes signature; // Authentication signature
     }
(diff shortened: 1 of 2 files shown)


service_contracts/src/PandoraService.sol
```diff
@@ -60,23 +67,28 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
     mapping(address => uint256) public clientDataSetIDs;
     // Mapping from proof set ID to root ID to metadata
     mapping(uint256 => mapping(uint256 => string)) public proofSetRootMetadata;
```
ZenGround0, line 69: I think you need to delete this now right?

ZenGround0 (commented):
Looking good but ill wait for the originators of #79 to weigh in before approving

service_contracts/src/PandoraService.sol
```diff
@@ -380,7 +392,20 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
         ProofSetInfo storage info = proofSetInfo[proofSetId];
         info.payer = createData.payer;
         info.payee = creator; // Using creator as the payee
-        info.metadata = createData.metadata;
+
+        // Store each metadata key-value entry for this proof set
+        for (uint256 i = 0; i < createData.metadataKeys.length; i++) {
```
rvagg, line 397:
```suggestion
        // Store each metadata key-value entry for this proof set
        require(createData.metadataKeys.length, createData.metadataValues, "Length mismatch between metadata keys and values");
        for (uint256 i = 0; i < createData.metadataKeys.length; i++) {
```

service_contracts/src/PandoraService.sol
```diff
@@ -380,7 +392,20 @@ contract PandoraService is PDPListener, IArbiter, Initializable, UUPSUpgradeable
...
+            string memory key = createData.metadataKeys[i];
+            bytes memory value = createData.metadataValues[i];
+
+            if (proofSetMetadata[proofSetId][key].length == 0) {
+                // If the key doesn't exist, add it to the keys array
+                proofSetMetadataKeys[proofSetId].push(key);
+            }
```
rvagg, line 404: In the `proofSetCreated` case, I think I'd prefer to just:


```suggestion
            require(proofSetMetadata[proofSetId][key].length == 0, "Duplicate metadata key provided");
```

Is there a good reason you can think of to allow duplicates for a new proof set?

service_contracts/src/PandoraService.sol
pali101: @rvagg  Since we’re storing strings on-chain, the incremental cost of the mapping itself is almost negligible compared to the cost of storing the string values. 

However, in our current design we end up duplicating these same key names in storage repeatedly for every proof set (e.g., `filename`, `version` etc). To optimize, I suggest we define a set of common keys (pre-defined), we can define them using either an enum or fixed bytes32 values. This way, for each proof set root, we only store the values for those keys, without needing to store the key name every time. For example:
```solidity
enum MetadataKey { Filename, Version, other common fields, Misc}
mapping(uint256 => mapping(uint8 => bytes)) public proofSetRootMetadata;
```
or
```solidity
mapping(uint256 => mapping(bytes32 => bytes)) public proofSetRootMetadata;
```
This approach makes the storage much more gas-efficient, since the keys are implied by their enum position or constant value, and aren’t repeatedly stored on-chain. Additionally, we can include a `misc` field among the predefined keys. This would act as a catch-all for any arbitrary or custom metadata that doesn’t fit the standard keys.

We should be able to find significant overlap across metadata keys. I’m unable to think of a use case where the metadata keys would always need to be fully dynamic.

service_contracts/src/PandoraService.sol
pali101: Updated

service_contracts/src/PandoraService.sol
pali101: Done

rjan90:
@rvagg does the latest changes here look okay to you?

rvagg:
I think we should put some basic limits in this for sizes, just to give an upper bound; we can be generous and let them pay the gas if they really want to push it:

```solidity
uint256 constant MAX_KEY_LENGTH = 64;
uint256 constant MAX_VALUE_LENGTH = 512;
uint256 constant MAX_KEYS_PER_PROOFSET = 10;
uint256 constant MAX_KEYS_PER_ROOT = 5;
```

rvagg:
Aside from the comments above (some of the Copilot ones are 👍) this is pretty good.

rvagg (approved):

rjan90:
@pali101 sorry for not getting this ⛙ before some conflicts popped up - do you want to address those, and then we can get this landed?

rvagg:
It's not simple conflicts btw, it's the comprehensive renaming, and you're going to have to do that renaming in here, `ProofSet` -> `DataSet`, `Root` -> `Piece` etc. Sorry we didn't get this in before the rename.

rvagg:
@pali101 would you mind getting back to this? I was in the SDK and noticed it wasn't there but it would be good to get this in ASAP with all the current breaking changes we're doing.

rjan90:
Closing this, as it has been superseded by: https://github.com/FilOzone/filecoin-services/pull/131

PR #87, opened 2025-07-23
feat: handle rail termination callback from payments contract

aarshkshah1992:
Now that we have a `railTerminated` call back from the payments contract:

Once a rail is terminated:

- SP can not add new roots to the proofset as we can't increase the rate of a terminated rail

- After the rail end epoch has elapsed (i.e. the rail "expires"):
   - SP can only delete the proofset or change ownership. No other operation on the proofset is allowed.

### DIFF
--- a/service_contracts/src/PandoraService.sol
+++ b/service_contracts/src/PandoraService.sol
@@ -30,6 +30,7 @@ contract PandoraService is PDPListener, IValidator, Initializable, UUPSUpgradeab
     event DataSetRailCreated(uint256 indexed dataSetId, uint256 railId, address payer, address payee, bool withCDN);
     event RailRateUpdated(uint256 indexed dataSetId, uint256 railId, uint256 newRate);
     event PieceMetadataAdded(uint256 indexed dataSetId, uint256 pieceId, string metadata);
+    event RailTerminatedInService(uint256 indexed railId, address indexed terminator, uint256 endEpoch);
 
     // Constants
     uint256 public constant NO_CHALLENGE_SCHEDULED = 0;
@@ -117,6 +118,15 @@ contract PandoraService is PDPListener, IValidator, Initializable, UUPSUpgradeab
 
     // Track when proving was first activated for each data set
     mapping(uint256 => uint256) public provingActivationEpoch;
+
+    // Rail termination tracking
+    struct RailTerminationStatus {
+        bool isTerminated;
+        uint256 endEpoch;
+    }
+
+    // Mapping from rail ID to termination status
+    mapping(uint256 => RailTerminationStatus) public railTerminationStatus;
 
     // ========== Storage Provider Registry State ==========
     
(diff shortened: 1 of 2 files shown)


service_contracts/src/PandoraService.sol
```diff
@@ -532,6 +545,13 @@ contract PandoraService is PDPListener, IValidator, Initializable, UUPSUpgradeab
             "Data set not registered with payment system"
         );
         
+        // Check if rail is terminated and beyond end epoch
+        RailTerminationStatus memory termStatus = railTerminationStatus[info.railId];
+        require(
+            !termStatus.isTerminated || block.number <= termStatus.endEpoch,
+            "Operation rejected: rail terminated and beyond end epoch"
+        );
```
juliangruber, line 553: can we refactor this into a method like `requireActivePaymentRail(railId)` or `requireUnterminatedPaymentRail(railId)`? It is being used without differences in 3 places

service_contracts/src/PandoraService.sol
```diff
@@ -594,6 +624,16 @@ contract PandoraService is PDPListener, IValidator, Initializable, UUPSUpgradeab
         external
         onlyPDPVerifier
     {
+        // Check if rail is terminated and beyond end epoch
+        DataSetInfo storage info = dataSetInfo[dataSetId];
+        if (info.railId != 0) {
+            RailTerminationStatus memory termStatus = railTerminationStatus[info.railId];
+            require(
+                !termStatus.isTerminated || block.number <= termStatus.endEpoch,
+                "Operation rejected: rail terminated and beyond end epoch"
```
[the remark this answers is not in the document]
juliangruber: I think we can remove the prefix `Operation rejected`, since user will already see in the transaction data that it has been required

service_contracts/src/PandoraService.sol
```diff
@@ -1356,4 +1396,39 @@ contract PandoraService is PDPListener, IValidator, Initializable, UUPSUpgradeab
             note: ""
         });
     }
+
+    /**
+     * @notice Called when a payment rail is terminated in the Payments contract
+     * @dev Implements the IValidator interface function
+     * @param railId ID of the payment rail being terminated
+     * @param terminator Address that initiated the termination
+     * @param endEpoch The final epoch up to which the rail can be settled
```
juliangruber, line 1405: I might not understand settling fully, but does this mean funds will be locked inside the rail if I didn't settle them before `endEpoch`?

service_contracts/src/PandoraService.sol
```diff
@@ -1356,4 +1396,39 @@ contract PandoraService is PDPListener, IValidator, Initializable, UUPSUpgradeab
...
+     */
+    function railTerminated(uint256 railId, address terminator, uint256 endEpoch) external override {
+        // Only payments contract can call this
+        require(msg.sender == paymentsContractAddress, "Only payments contract can terminate rails");
```
juliangruber, line 1409: since only the one payments contract can do this, shall we drop the `terminator` argument?

juliangruber (changes requested):

service_contracts/src/PandoraService.sol
juliangruber: I can add termination logic for the FilCDN rails when we have decided on this matter

service_contracts/src/PandoraService.sol
juliangruber: If that is the case then the error message in the `require()` is confusing

aarshkshah1992:
Closing this in favour of "https://github.com/FilOzone/filecoin-services/pull/93".

PR #100, opened 2025-07-29
Private Constants

wjmelements:
Reviewer @pali101
This brings the contract size below 24k.
#### Changes
* privatize all constants

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -27,7 +27,7 @@ contract FilecoinWarmStorageService is
     EIP712Upgradeable
 {
     // Version tracking
-    string public constant VERSION = "0.1.0";
+    string private constant VERSION = "0.1.0";
 
     // Events
     event ContractUpgraded(string version, address implementation);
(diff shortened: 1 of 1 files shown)


wjmelements:
I think some of the fields set by `initialize` could be `immutable`. That is a harder change because some logic has to move to the constructor and both testing and deployment code needs to be updated. But reducing the size of `initialize` should also reduce runtime codesize.

wjmelements:
The immutable change reduces it another 569 , to `27348`. The total runtime codesize reduction of this changeset is now `1067`.

wjmelements:
> The immutable change reduces it another 569 , to `27348`. The total runtime codesize reduction of this changeset is now `1067`.

looks like I pushed this to [a different branch](https://github.com/FilOzone/filecoin-services/compare/main...wjmelements:filecoin-services:immutable-fields). Gonna rebase those changes onto this.

The size after [f18b83f](https://github.com/FilOzone/filecoin-services/pull/100/commits/f18b83fc6dcb5ef589e39359804a731645b14483) is 27874 (-43).

The size after [dbcdc0c](https://github.com/FilOzone/filecoin-services/pull/100/commits/dbcdc0c8bd19c2edfa46c05a87243622f24cea59) is 27863 (-11).

wjmelements:
Size after 0d33830 is `27294` (-569). Total codesize reduction from this changeset is `27294`.

rvagg:
seems fine to me, I just checked in the SDK and we're already using public accessors to get the properties we care about that are being privatised here

pali101:
Great refactor! We should be under `24KB` limit now. 

A small suggestion: since these values are now `private`, other contracts would have a hard time fetching them directly. While it’s technically possible to use an oracle and read logs from blockchain data to extract these values, this is not practical or efficient for on-chain interactions.

For transparency and easier on-chain integration, could we add grouped public getter functions for the main fee constants and contract addresses? This would allow SDKs and other contracts to fetch all relevant config values with a single call, even if the variables are private. 

Implementing grouped getters shouldn't significantly increase the runtime size compared to having individual getters.

Here’s an example for reference returning values as tuple (or a single struct if preferred):

```solidity
function getFeeConstants()
    public
    view
    returns (
        uint256 spRegistrationFee,
        uint256 dataSetCreationFee,
        uint256 storagePricePerTiBPerMonth,
        uint256 cacheMissPricePerTiBPerMonth,
        uint256 cdnPricePerTiBPerMonth
    )
{
    spRegistrationFee = SP_REGISTRATION_FEE;
    dataSetCreationFee = DATA_SET_CREATION_FEE;
    storagePricePerTiBPerMonth = STORAGE_PRICE_PER_TIB_PER_MONTH;
    cacheMissPricePerTiBPerMonth = CACHE_MISS_PRICE_PER_TIB_PER_MONTH;
    cdnPricePerTiBPerMonth = CDN_PRICE_PER_TIB_PER_MONTH;
}


function getImportantAddresses()
    public
    view
    returns (address pdpVerifier, address payments, address usdfc, address filCDN)
{
    pdpVerifier = pdpVerifierAddress;
    payments = paymentsContractAddress;
    usdfc = usdfcTokenAddress;
    filCDN = filCDNAddress;
}


function getVersion() public pure returns (string memory) {
    return VERSION;
}
```

aarshkshah1992 (approved):
@wjmelements Please can you push a forge fmt change and we can then merge this. Nice refactor.

PR #101, opened 2025-07-29
initializeV2 -> configureProvingPeriod

wjmelements:
Reviewer @aarshkshah1992
It is true that upgrades should not call initialize again, but the upgraded implementations should have their own initialize functions.
State variables like `maxProvingPeriod` and `challengeWindowSize` can be configured by `onlyOwner` functions instead of by initialize.
So I refactor `initializeV2` into `configureProvingPeriod`.
I do not know the risks of configuring this proving period, but it just as possible as before.
#### Changes
* rename method, change modifiers
* fix tests
* rm initializeV2 script


aarshkshah1992 (approved):

ZenGround0:
I'm not sure I like this.  We don't want to reconfigure proving period parameters we only want to set them on initialization.  Isn't initializeV2 a good new initialization function for the upgraded implementations?  I'm missing the problem that this solved.

> I do not know the risks of configuring this proving period, but it just as possible as before.

I think you are saying this is because the contract can be upgraded to new values by the owner, right?

PR #108, opened 2025-07-31
feat!: add name and description

jennijuju:
Adding name and description fields, and emitting a deployment event, will make explorer indexing and integrations easier in the long term if standardized for Filecoin service contracts in the early stages.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -35,6 +35,7 @@ contract FilecoinWarmStorageService is
     event DataSetStorageProviderChanged(
         uint256 indexed dataSetId, address indexed oldStorageProvider, address indexed newStorageProvider
     );
+    event ServiceDeployed(string name, string description, address indexed deployedAt);
     event FaultRecord(uint256 indexed dataSetId, uint256 periodsFaulted, uint256 deadline);
     event DataSetRailsCreated(
         uint256 indexed dataSetId,
@@ -280,6 +281,9 @@ contract FilecoinWarmStorageService is
         name = "Filecoin Warm Storage Service";
         // TODO: add the url w/ more service info
         description = "Verifiable storage powered by Filecoin PDP, with optional CDN integration for fast content delivery.";
+
+        // Emit deployment event
+        emit ServiceDeployed(name, description, address(this));
     }
 
     function _authorizeUpgrade(address newImplementation) internal override onlyOwner {}
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -173,6 +173,10 @@ contract FilecoinWarmStorageService is
     uint64 private maxProvingPeriod;
     uint256 private challengeWindowSize;
 
+    // Service metadata
+    string public name;
+    string public description;
+
     // Events for SP registry
     event ProviderRegistered(address indexed provider, string serviceURL, bytes peerId);
     event ProviderApproved(address indexed provider, uint256 indexed providerId);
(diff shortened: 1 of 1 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -271,6 +276,15 @@ contract FilecoinWarmStorageService is
...
         nextServiceProviderId = 1;
+
+        // Initialize service metadata
+        name = "Filecoin Warm Storage Service";
+        // TODO: add the url w/ more service info
+        description =
+            "Verifiable storage powered by Filecoin PDP, with optional CDN integration for fast content delivery.";
+
+        // Emit deployment event
+        emit ServiceDeployed(name, description, address(this));
```
wjmelements, line 287: address(this) is part of every log by default

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -271,6 +276,15 @@ contract FilecoinWarmStorageService is
         cdnServiceCommissionBps = 0; // 0%
 
         nextServiceProviderId = 1;
+
+        // Initialize service metadata
+        name = "Filecoin Warm Storage Service";
+        // TODO: add the url w/ more service info
+        description =
```
wjmelements, line 283: a more typical way of providing this information is to put it in the comments of the verified contract code in the block explorer.

The name of the contract is also part of the contract source verification.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -173,6 +174,10 @@ contract FilecoinWarmStorageService is
     uint64 private maxProvingPeriod;
     uint256 private challengeWindowSize;
 
+    // Service metadata
+    string public name;
```
wjmelements, line 178: this can be constant

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -322,6 +336,15 @@ contract FilecoinWarmStorageService is
         cdnServiceCommissionBps = newCDNCommissionBps;
     }
 
+    /**
+     * @notice Updates the service description
+     * @dev Only callable by the contract owner
+     * @param newDescription The new description for the service
+     */
+    function updateDescription(string calldata newDescription) external onlyOwner {
```
wjmelements, line 344: Is this likely to be invoked? I believe this contract is bumping into the codesize limit.

wjmelements:
Events only emitted by the constructor are fairly harmless. They would increase initcodesize but not codesize.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: Which explorer do you have in mind? What do their events look like, or where do you expect for this to be surfaced?

wjmelements:
If we use delegatecall proxies, `name`, `description`, and `version` can be public constants. This would mean they don't have to be updated in `migrate`/`initialize`, but they would be automatically updated when we switched to a new implementation.

I believe the current plan is to remove the proxy layer, in which case these fields should still be constant.

rvagg:
From sync chat just now: it would be good to have this enforced by the payments contract - any service it interacts with should be identifiable in a standardised way, to allow for deeper inspection starting at the payments contract and sifting down into services that wallet is using as validators for payments.

wjmelements:
> any service it interacts with should be identifiable in a standardised way

I don't think there's a standardized self-identification interface. Closest thing I know of is ERC20 metadata. One problem with self-identification is that you can lie; for example there are 506 "USDT" tokens on Ethereum mainnet last I checked.

PR #80, opened 2025-07-11
Beneficiary Separation: Hot/Cold Key Architecture

Hany-Almnaem:
##  **Overview**

This PR implements **beneficiary separation** to enable hot/cold key architecture for storage providers, allowing separation of control operations and payment receiving addresses while maintaining full backward compatibility.

### **Key Separation Model**
- **Hot Key (Control Address)**: Used for proof set operations, stored as `payee`
- **Cold Key (Beneficiary Address)**: Used for payment receiving, routed through payment rails
- **Backward Compatibility**: Existing proof sets continue working unchanged

### **Payment Rail Routing**
- Before: Payer → Control Address (creator)
- After:  Payer → Beneficiary Address (cold key)
- Legacy: Payer → Control Address (fallback)

### **Format Detection**
- **Legacy Format (4 params)**: `(metadata, payer, withCDN, signature)`
- **New Format (5 params)**: `(metadata, payer, withCDN, beneficiary, signature)`
- **Robust Detection**: Try/catch mechanism for seamless format handling

### **Signature Security**
- **Beneficiary NOT included** in client signatures (SP-controlled)
- **Replay protection** via clientDataSetId
- **EIP-712 compliance** maintained

### **Test Coverage**
- Comprehensive Test Suite (68/68 passing)
- Legacy format support with proper fallback
- New format validation with beneficiary checks
- Mixed scenario testing (legacy + new in same system)
- Event verification for both formats
- Gas optimization 
- Edge case handling (zero addresses, malformed data)
- Integration testing with existing functionality

### **🔗 Related Issues**
Close #68

PR #111, opened 2025-08-04
Feature/beneficiary separation new

### LINKED ISSUE #68: Separate Beneficiary Address
From an SPs perspective there are two distinct ownership roles for a private key involved in the pdp service 
1. A control address that owns a proofset and can submit operations 
2. A beneficiary address that receives payment for providing a service 

These keys will be handled differently (roughly hot/ cold) and should be separated out.

To do this concretely we should 
* Add a beneficiary address field to the create proofset extra data and parse this out.  This should NOT be signed by the client i.e. `decodeProofSetCreateData` should return `ProofSetCreateData` and `beneficiary` separately and no modification needed to `verifyCreateProofSetSignature`
* Create payment rails between payer and beneficiary no longer payer and creator 
* creator can still be listed as payee on the proofset info: it is a more useful address for client to reference + anyone can go to the payments contract with the proofset info's railid if they really want to lookup the beneficiary address.

Follow up 
* Propagate this change to curio by adding in a beneficiary address to the extradata before submitting message to chain

Hany-Almnaem:
### **Overview**
This PR implements a secure beneficiary separation feature that maintains backward compatibility while providing enhanced security for new datasets.

#### **Contract Changes**
- **Added `isNewFormat` field** to `DataSetInfo` struct for explicit format tracking
- **Enhanced `dataSetCreated`** to set format flag and handle both legacy (4-parameter) and new (5-parameter) formats
- **Updated `storageProviderChanged`** with secure logic:
  - Legacy datasets: beneficiary follows payee (preserves existing behavior)
  - New format datasets: beneficiary never changes (hot/cold key separation)
- **Maintained backward compatibility** with existing legacy datasets

#### **Security Model**
| Format | Creation | SP Change | Security Guarantee |
|--------|----------|-----------|-------------------|
| **Legacy (4p)** | `beneficiary = creator` | `beneficiary` follows `payee` | Legacy behavior preserved |
| **New (5p)** | `beneficiary = explicit` | `beneficiary` never changes | Hot/cold key separation |

#### **Test Coverage**
-  Legacy format storage provider changes
- New format storage provider changes  
- Multiple datasets per provider scenarios
- Arbitrary extra data handling
- All existing functionality preserved

### **Security Benefits**
1. **Explicit format tracking** eliminates ambiguity
2. **Deterministic security model** prevents edge case attacks
3. **Hot/cold key separation** for new format datasets
4. **No breaking changes** to existing functionality

#### **Legacy Format (4 parameters)**
```solidity
// Dataset creation
info.beneficiary = creator; // Initial storage provider
info.isNewFormat = false;

// Storage provider change
if (!info.isNewFormat && info.beneficiary == oldStorageProvider) {
    info.beneficiary = newStorageProvider; // Follows payee
}
```

#### **New Format (5 parameters)**
```solidity
// Dataset creation
info.beneficiary = explicitBeneficiary; // User-specified
info.isNewFormat = true;

// Storage provider change
// beneficiary never changes (hot/cold separation)
```

### **Testing**
- All 63 tests pass
- Comprehensive coverage of both formats
- Edge cases handled securely
- Backward compatibility verified

### **Breaking Changes**
None. This is a purely additive feature that maintains full backward compatibility.

---

Closes #68 
Supersedes #80

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -32,8 +32,8 @@ contract FilecoinWarmStorageService is
 
     // Events
     event ContractUpgraded(string version, address implementation);
-    event DataSetserviceProviderChanged(
-        uint256 indexed dataSetId, address indexed oldserviceProvider, address indexed newserviceProvider
+    event DataSetServiceProviderChanged(
+        uint256 indexed dataSetId, address indexed oldServiceProvider, address indexed newServiceProvider
     );
     event FaultRecord(uint256 indexed dataSetId, uint256 periodsFaulted, uint256 deadline);
     event DataSetRailsCreated(
@@ -146,7 +146,7 @@ contract FilecoinWarmStorageService is
 
     // ========== Service Provider Registry State ==========
 
-    uint256 private nextserviceProviderId = 1;
+    uint256 private nextServiceProviderId = 1;
 
     struct ApprovedProviderInfo {
         address serviceProvider;
@@ -260,7 +260,7 @@ contract FilecoinWarmStorageService is
         // Set commission rate
         serviceCommissionBps = 0; // 0%
 
-        nextserviceProviderId = 1;
+        nextServiceProviderId = 1;
     }
 
     function _authorizeUpgrade(address newImplementation) internal override onlyOwner {}
@@ -356,13 +356,13 @@ contract FilecoinWarmStorageService is
     // Getters
     function getAllApprovedProviders() external view returns (ApprovedProviderInfo[] memory) {
         // Handle edge case: no providers have been registered
-        if (nextserviceProviderId == 1) {
+        if (nextServiceProviderId == 1) {
             return new ApprovedProviderInfo[](0);
         }
 
         // First pass: Count non-empty providers (those with non-zero service provider address)
         uint256 activeCount = 0;
-        for (uint256 i = 1; i < nextserviceProviderId; i++) {
+        for (uint256 i = 1; i < nextServiceProviderId; i++) {
             if (approvedProviders[i].serviceProvider != address(0)) {
                 activeCount++;
             }
@@ -378,7 +378,7 @@ contract FilecoinWarmStorageService is
 
         // Second pass: Fill array with only active providers
         uint256 currentIndex = 0;
-        for (uint256 i = 1; i < nextserviceProviderId; i++) {
+        for (uint256 i = 1; i < nextServiceProviderId; i++) {
             if (approvedProviders[i].serviceProvider != address(0)) {
                 providers[currentIndex] = approvedProviders[i];
                 currentIndex++;
@@ -408,7 +408,7 @@ contract FilecoinWarmStorageService is
         require(creator != address(0), Errors.ZeroAddress(Errors.AddressField.Creator));
 
         // Check if the service provider is whitelisted
-        require(approvedProvidersMap[creator], Errors.serviceProviderNotApproved(creator));
+        require(approvedProvidersMap[creator], Errors.ServiceProviderNotApproved(creator));
 
         // Update client state
         uint256 clientDataSetId = clientDataSetIDs[createData.payer]++;
@@ -428,6 +428,8 @@ contract FilecoinWarmStorageService is
         info.commissionBps = serviceCommissionBps;
         info.clientDataSetId = clientDataSetId;
         info.withCDN = createData.withCDN;
+
+        // Note: The payer must have pre-approved this contract to spend USDFC tokens before creating the data set
 
         // Create the payment rails using the Payments contract
         Payments payments = Payments(paymentsContractAddress);
@@ -484,7 +486,7 @@ contract FilecoinWarmStorageService is
                 filCDNAddress,
                 address(this), // arbiter
                 0, // no service commission
-                address(this) // terminator
+                address(this)
             );
             info.cdnRailId = cdnRailId;
             railToDataSet[cdnRailId] = dataSetId;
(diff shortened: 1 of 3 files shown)


rjan90:
Hey @Hany-Almnaem! If you rebase your PR on top of the latest changes on main, the contract size check should pass, as a couple of size optimisations landed here: https://github.com/FilOzone/filecoin-services/pull/110

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -94,13 +97,15 @@ contract FilecoinWarmStorageService is
...
-        address payee; // SP's beneficiary address
+        address payee; // SP's control address (hot key)
+        address beneficiary; // SP's payment address (cold key)
         uint256 commissionBps; // Commission rate for this data set (dynamic based on whether the client purchases CDN add-on)
         string metadata; // General metadata for the data set
         string[] pieceMetadata; // Array of metadata for each piece
         uint256 clientDataSetId; // ClientDataSetID
         bool withCDN; // Whether the data set is registered for CDN add-on
         uint256 paymentEndEpoch; // 0 if payment is not terminated
+        bool isNewFormat; // Track whether this dataset uses new format (explicit beneficiary)
```
ZenGround0, line 108: We don't want to maintain the old format.  Let's remove this boolean and enforce that the new format is used.

ZenGround0: A bit more context: because we are still in development we are throwing away old contracts and moving over to totally new deployments somewhat frequently.  While we are free of being forced to maintain old systems we should take the opportunity to keep the initial system we launch as clean as possible.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -415,46 +434,32 @@ contract FilecoinWarmStorageService is
         // Initialize the DataSetInfo struct
         DataSetInfo storage info = dataSetInfo[dataSetId];
         info.payer = createData.payer;
-        info.payee = creator; // Using creator as the payee
+        info.payee = creator; // Hot key for control operations
+        info.beneficiary = actualBeneficiary; // Cold key for payments - ADDED THIS
```
ZenGround0, line 438: Beneficiary means receiver of funds so payee should be called beneficiary and info.payee should be renamed to info.controller

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -415,46 +434,32 @@ contract FilecoinWarmStorageService is
...
         // Create the payment rails using the Payments contract
         Payments payments = Payments(paymentsContractAddress);
         uint256 pdpRailId = payments.createRail(
             usdfcTokenAddress, // token address
             createData.payer, // from (payer)
-            creator, // data set creator, SPs in  most cases
-            address(this), // this contract acts as the validator
-            info.commissionBps, // commission rate based on CDN usage
-            address(this)
+            actualBeneficiary, // to (beneficiary) - CHANGED from creator
```
ZenGround0, line 450: In general please remove the agent comments which are helpful in development when looking at the diff but not useful for maintaining the code.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -415,46 +434,32 @@ contract FilecoinWarmStorageService is
...
         );
 
         // Store the rail ID
         info.pdpRailId = pdpRailId;
-
-        // Store reverse mapping from rail ID to data set ID for validation
         railToDataSet[pdpRailId] = dataSetId;
 
-        // First, set a lockupFixed value that's at least equal to the one-time payment
-        // This is necessary because modifyRailPayment requires that lockupFixed >= oneTimePayment
```
ZenGround0, line 444: All the comments here are useful so they shouldn't be deleted.  I would probably revert all diff from 439 to 462.

ZenGround0 (changes requested):

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -394,7 +399,11 @@ contract FilecoinWarmStorageService is
     function dataSetCreated(uint256 dataSetId, address creator, bytes calldata extraData) external onlyPDPVerifier {
         // Decode the extra data to get the metadata, payer address, and signature
         require(extraData.length > 0, Errors.ExtraDataRequired());
-        DataSetCreateData memory createData = decodeDataSetCreateData(extraData);
+
+        // Decode new format only: metadata, payer, withCDN, beneficiary, signature
```
ZenGround0, line 403: Please remove the AI descriptive comments.  They are useful in the context of you reviewing the model's output before pushing.  They are not useful in the context of creating clear maintainable code.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -409,29 +418,28 @@ contract FilecoinWarmStorageService is
...
         Payments payments = Payments(paymentsContractAddress);
         uint256 pdpRailId = payments.createRail(
             usdfcTokenAddress, // token address
             createData.payer, // from (payer)
-            creator, // data set creator, SPs in  most cases
+            actualBeneficiary, // to (beneficiary)
             address(this), // this contract acts as the validator
             info.commissionBps, // commission rate based on CDN usage
-            address(this)
+            address(this) // terminator
```
ZenGround0, line 442: Remove this comment since it is inaccurate.  This is the service fee recipient, payment rails don't have a terminator concept.  I am guessing this comment was AI generated.  Please remember that models are wrong very often when commenting.  When you see a newly introduced comment which is not very related to the requested change you should take the time to check if it makes sense.  This is also a good way to learn more about the system you're contributing to.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -463,10 +471,10 @@ contract FilecoinWarmStorageService is
             cacheMissRailId = payments.createRail(
                 usdfcTokenAddress, // token address
                 createData.payer, // from (payer)
-                creator, // data set creator, SPs in most cases
-                address(this), // this contract acts as the arbiter
+                actualBeneficiary,
+                address(this), // arbiter
                 0, // no service commission
-                address(this)
+                address(this) // terminator
```
ZenGround0, line 477: ditto

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -476,19 +484,22 @@ contract FilecoinWarmStorageService is
                 usdfcTokenAddress, // token address
                 createData.payer, // from (payer)
                 filCDNAddress,
-                address(this), // this contract acts as the arbiter
+                address(this), // arbiter
                 0, // no service commission
-                address(this)
+                address(this) // terminator
```
ZenGround0, line 489: ditto

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -476,19 +484,22 @@ contract FilecoinWarmStorageService is
...
             payments.modifyRailLockup(cdnRailId, DEFAULT_LOCKUP_PERIOD, 0);
         }
 
-        // Emit event for tracking
+        // Emit events for tracking
         emit DataSetRailsCreated(
             dataSetId, pdpRailId, cacheMissRailId, cdnRailId, createData.payer, creator, createData.withCDN
         );
+
+        // Emit beneficiary resolution event
```
ZenGround0, line 501: This is an example of a comment that does not benefit us because it is redundant with the code.  People reading the code will get almost nothing out of this comment which reduces the overall signal to noise ratio of comments and makes us all less likely to read helpful comments.

Please take a look at [Ousterhout's PhOSD](https://github.com/alysivji/notes/blob/main/software-engineering/philosophy_of_software_design.md#chapter-13-comments-should-describe-things-that-arent-obvious-from-the-code).  If you read the actual book cover to cover it will be generally very helpful.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -721,15 +732,18 @@ contract FilecoinWarmStorageService is
...
         // New service provider must be an approved provider
         require(approvedProvidersMap[newServiceProvider], Errors.NewServiceProviderNotApproved(newServiceProvider));
 
-        // Update the data set payee (service provider)
-        info.payee = newServiceProvider;
+        // Update the data set controller (storage provider)
+        info.controller = newStorageProvider;
+
+        // Beneficiary is independent and should not auto-follow controller changes
+        // Beneficiary is independent in the new model, do not auto-update
```
ZenGround0, line 746: nit: keep only one line

ZenGround0 (changes requested):

service_contracts/src/FilecoinWarmStorageService.sol
ZenGround0: Looks like I said this last time too.  https://github.com/FilOzone/filecoin-services/pull/111#discussion_r2258224607

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -942,15 +956,38 @@ contract FilecoinWarmStorageService is
...
-     * @param extraData The encoded extra data from PDPVerifier
-     * @return decoded The decoded DataSetCreateData struct
+     * @notice Helper function to try decoding new format
+     * @dev This needs to be external for try/catch to work
      */
-    function decodeDataSetCreateData(bytes calldata extraData) internal pure returns (DataSetCreateData memory) {
-        (string memory metadata, address payer, bool withCDN, bytes memory signature) =
-            abi.decode(extraData, (string, address, bool, bytes));
+    // New-format decode helper
+    function tryDecodeNewFormat(bytes calldata extraData)
```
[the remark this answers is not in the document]
Hany-Almnaem: The external self-call and the "try/catch" frame should be removed, since we are implementing a single format.
But I'm currently having trouble merging, as I'm having trouble mixing old and new code. I'm working on that now. Sorry for the inconvenience

PR #114, opened 2025-08-05
ci: Add CI workflow to verify ABI consistency

pali101:
## Summary

This PR adds a GitHub Actions workflow that checks if smart contract ABIs in `subgraph/abis/` match the Solidity source. It regenerates ABIs using `make abi` and fails if differences with committed files are found, ensuring ABI updates accompany contract changes and preventing integration issues.

## Changes

- added a new GitHub Actions workflow Check ABI consistency
- added a make target `abi` that generates ABIs for all contracts in service_contracts/src (excluding libraries)
- added a script `service_contracts/tools/generate-abi.sh` used both in CI and the Makefile for ABI generation

### DIFF
--- a/.github/workflows/check-abi-consistency.yml
+++ b/.github/workflows/check-abi-consistency.yml
@@ -31,4 +31,4 @@ jobs:
         run: |
           cd service_contracts
           make abi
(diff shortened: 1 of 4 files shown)


ZenGround0:
@hugomrdias I think this will resolve your issue -- at least with this contract.

PR #115, opened 2025-08-06
Migration Preparation

ZenGround0:
This is a draft of the pieces of [this design](https://www.notion.so/filecoindev/Filecoin-Services-Model-for-Change-241dc41950c180af85f1cac1687766a5) that need to land right away.

Some open considerations and TODOs
- should we immediately free up lockup and locked funds?
- should we be communicating any information about the proof failure history of this dataset over migration
- we should fail on a terminated but non finalized dataset -- such a dataset should finish its life on the v1 contract


ZenGround0:
We are somewhat in limbo with the upgradability strategy on fwss contract.  Today we are still using `UUPSUpgradeable` contract so this does not apply.  We can come back to this if we decide to deploy a static solution, though we'll probably want to rethink the implementation as the contract has changed significantly.  Closing for now.

PR #127, opened 2025-08-08
fix: make deploy scripts match current contracts

rvagg:
Some simple updates to match current constructor and other changes in the contracts. Thankfully too many constructor arguments isn't a fatal thing in `forge create`, it'll just ignore them, so this goes unnoticed mainly.

### DIFF
--- a/service_contracts/tools/deploy-warm-storage-implementation-only.sh
+++ b/service_contracts/tools/deploy-warm-storage-implementation-only.sh
@@ -25,9 +25,33 @@ echo "Deploying from address: $ADDR"
 # Get current nonce
 NONCE="$(cast nonce --rpc-url "$RPC_URL" "$ADDR")"
 
+# Get required addresses from environment or use defaults
+if [ -z "$PDP_VERIFIER_ADDRESS" ]; then
+  echo "Error: PDP_VERIFIER_ADDRESS is not set"
+  exit 1
+fi
+
+if [ -z "$PAYMENTS_CONTRACT_ADDRESS" ]; then
+  echo "Error: PAYMENTS_CONTRACT_ADDRESS is not set"
+  exit 1
+fi
+
+if [ -z "$FILCDN_ADDRESS" ]; then
+  echo "Warning: FILCDN_ADDRESS not set, using default"
+  FILCDN_ADDRESS="0xff0000000000000000000000000000000002870c"
+fi
+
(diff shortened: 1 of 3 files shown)


service_contracts/tools/deploy-warm-storage-implementation-only.sh
```diff
@@ -25,9 +25,33 @@ echo "Deploying from address: $ADDR"
...
 # Deploy FilecoinWarmStorageService implementation
 echo "Deploying FilecoinWarmStorageService implementation..."
-WARM_STORAGE_IMPLEMENTATION_ADDRESS=$(forge create --rpc-url "$RPC_URL" --keystore "$KEYSTORE" --password "$PASSWORD" --broadcast --nonce $NONCE --chain-id 314159 src/FilecoinWarmStorageService.sol:FilecoinWarmStorageService --optimizer-runs 1 --via-ir | grep "Deployed to" | awk '{print $3}')
+echo "Constructor arguments:"
+echo "  PDPVerifier: $PDP_VERIFIER_ADDRESS"
+echo "  Payments: $PAYMENTS_CONTRACT_ADDRESS"
+echo "  USDFC Token: $USDFC_TOKEN_ADDRESS"
+echo "  FilCDN: $FILCDN_ADDRESS"
+
+WARM_STORAGE_IMPLEMENTATION_ADDRESS=$(forge create --rpc-url "$RPC_URL" --keystore "$KEYSTORE" --password "$PASSWORD" --broadcast --nonce $NONCE --chain-id 314159 src/FilecoinWarmStorageService.sol:FilecoinWarmStorageService --constructor-args $PDP_VERIFIER_ADDRESS $PAYMENTS_CONTRACT_ADDRESS $USDFC_TOKEN_ADDRESS $FILCDN_ADDRESS --optimizer-runs 1 --via-ir | grep "Deployed to" | awk '{print $3}')
```
rjan90, line 54: Just wondering if the `--optimizer-runs` flag should be 200 here to minimize bytecode size?

rvagg: yes, removed these flags because they should come from foundry.toml

rvagg:
Also added a check for challenge finality setting so that it makes sense; if set wrong, Curio will have trouble picking the right epoch (discovered in practice)

rjan90 (approved):

PR #128, opened 2025-08-08
refactor: sync subgraph code with renamed contract details

silent-cipher:
Syncing subgraph with renamed contract details ( PDP & FilecoinWarmStorageService ). 
Support for CommPv2 added


ZenGround0:
Are you planning to check in the newer abi jsons? I know that @hugomrdias has been relying on these recently.  If the plan is to move these json files somewhere else (i.e. to their own respective contracts) that's fine we just need to be clear about it.

silent-cipher:
I’m keeping only the ABI parts needed for the subgraph. Removed unused events/functions so we don’t have to update the ABI if other functions change.

hugomrdias:
yeah i need the full abi, should we move it to the repos ?

subgraph/src/utils/cid.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export class UvarintResult {
+  constructor(
+    public isValid: boolean,
+    public value: BigInt = BigInt.zero(),
+    public offset: BigInt = BigInt.zero()
+  ) {}
+}
+
+export function readUvarint(data: Bytes, offset: BigInt): UvarintResult {
```
rvagg, line 22: pretty rough that we have to implement uvarint parsing in here, do we have no option for dependencies in these things?

silent-cipher: implementing uvarint parsing in here just to get size from cid- there's no way we can use standard libraries. We need dedicated Assembly Script dependencies

Kubuxu:
I'm going to merge this to continue working on it.

Issue #95, opened 2025-07-28
terminateDataSetPayment


jennijuju:
A couple of thoughts on https://github.com/FilOzone/filecoin-services/pull/93:
- I think the external function should be `terminateService(dataset ID, extraData) where terminateDataSetPayment is an internal function it calls
   - in the future we might have `terminateService(dealID, extraData)` if there is a real deal protocol..
- terminateService should take a couple steps more then just terminating the payment rail imho, i.e: it should also trigger deletionDataSet so that the verifier is also informed and will not keep expecting that proof; likely the extraData should include some client signature

pyropy:
I assume that this method would also terminate cache-miss and cache-hit payment rails if payer has enabled CDN service for the data set? 

Also we should add a separate method for terminating and enabling CDN service for a data set but that outside of the scope for this issue.

Issue #96, opened 2025-07-29
Core logic missing in dataSetDeleted and piecesScheduledRemove


pali101:
The functions `dataSetDeleted` and `piecesScheduledRemove` currently perform only verification and do not implement their core logic:

- `dataSetDeleted`
https://github.com/FilOzone/filecoin-services/blob/2c45fb0fd51f3355fcf134ffe403d73b7ce31226/service_contracts/src/FilecoinWarmStorageService.sol#L518
- `piecesScheduledRemove`
https://github.com/FilOzone/filecoin-services/blob/2c45fb0fd51f3355fcf134ffe403d73b7ce31226/service_contracts/src/FilecoinWarmStorageService.sol#L582

Both functions need implementation of their intended state updates.

rjan90:
@aarshkshah1992 this should be in place for M2 right, so that we can properly exercise rail termination?

aarshkshah1992:
Now for `dataSetDeleted ` -> we expect clients and SPs to call `terminateDataSet` directly on the contract to terminate a rail. There are currently discussions to delete a data set as part of `terminateDataSet` or vice versa to make this cleaner.

Issue #125, opened 2025-08-08
Add ability to toggle CDN service for a dataset


pyropy:
> I assume that this method would also terminate cache-miss and cache-hit payment rails if payer has enabled CDN service for the data set? 
> 
> Also we should add a separate method for terminating and enabling CDN service for a data set but that outside of the scope for this issue. 

 _Originally posted by @pyropy in [#95](https://github.com/FilOzone/filecoin-services/issues/95#issuecomment-3166917424)_

**Description:**
We need to introduce functionality for toggling CDN service for a given dataset. This should be implemented as a **separate method** from the existing payment rail termination flow.

**Key points:**

* Toggling should be **per dataset**.
* Only the **dataset owner (client)** should be authorized to enable or disable CDN service.

**Acceptance criteria:**

* [ ] New method to toggle CDN service for a dataset.
* [ ] Authorization checks to ensure only the dataset owner can call these methods.

aarshkshah1992:
@pyropy You can "freeze" a rail by setting it's rate to 0 then change it back to non-0 whe you wanna start it again. Does that make sense ?

aarshkshah1992:
@pyropy How important is this for now ? Currently, the contract assumes that all the rails are terminated together and has only one end epoch across all of them. If we do this issues , that will no longer be true and we'll have to track seperate end epochs for all the rails and update our functions and callbacks accordingly.

jennijuju:
Ha thanks for creating this issue - it’s been on my backlog forever

jennijuju:
actually i think this is duped with https://github.com/FilOzone/filecoin-services/issues/75 (found it when i revisted last convo we had on this topic https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1751504874953929

pyropy:
@jennijuju I'm going to close of the issue as duplicate and continue the conversation there.

/cc @aarshkshah1992

PR #129, opened 2025-08-08
feat: rename datset termination and emit events for dataset termination and add extradata

aarshkshah1992:
Closes https://github.com/FilOzone/filecoin-services/issues/95 AND https://github.com/FilOzone/filecoin-services/issues/96.

- Once a service for a dataset is terminated, all it's payment rails are terminated. After this, SP can only remove pieces from a dataset upto the rail end epoch. Adding pieces will fail after a data set is terminated.
- SP can continue submitting proofs for a dataset after temrination all the way to `endEpoch` to continue to get paid for epochs between [termination epoch, end epoch].
 - After the rail end epoch, all calls including submitting proofs and removing pieces will ALSO revert with a meaningful error message asking the SP to delete the proofset on the `PDPVerifier`.
 - We also emit meaningful events here for dataset and rail termination.
 
 Note that SP is responsible for calling `deleteProofset` on the `PDPVerifier` after a dataset service is beyond it's `endEpoch`. It can either listen to dataset termination events or rail termination events to schedule this.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -48,6 +48,14 @@ contract FilecoinWarmStorageService is
     event RailRateUpdated(uint256 indexed dataSetId, uint256 railId, uint256 newRate);
     event PieceMetadataAdded(uint256 indexed dataSetId, uint256 pieceId, string metadata);
 
+    event ServiceTerminated(
+        address indexed caller, uint256 indexed dataSetId, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
+    );
+
+    event PaymentTerminated(
+        uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
+    );
+
     // Constants
     uint256 private constant NO_CHALLENGE_SCHEDULED = 0;
     uint256 private constant CHALLENGES_PER_PROOF = 5;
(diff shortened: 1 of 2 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -750,6 +758,8 @@ contract FilecoinWarmStorageService is
             payments.terminateRail(info.cacheMissRailId);
             payments.terminateRail(info.cdnRailId);
         }
+
+        emit ServiceTerminated(msg.sender, dataSetId, info.pdpRailId, info.cacheMissRailId, info.cdnRailId);
```
pyropy, line 762: Should this event also be emitted in case of the cdn toggle functionality we've discussed in #125 ? If so I assume we'd remove `pdpRailId` from the emitted event?

aarshkshah1992:
Let's wait for @ZenGround0 to chime in.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -729,7 +737,7 @@ contract FilecoinWarmStorageService is
         emit DataSetStorageProviderChanged(dataSetId, oldStorageProvider, newStorageProvider);
     }
 
-    function terminateDataSetPayment(uint256 dataSetId) external {
+    function terminateService(uint256 dataSetId, bytes calldata /*extraData*/ ) external {
```
ZenGround0, line 740: What's the point of this extra data?  At first I thought this was the callback from payments but that's not the case, it looks like this is just a top level method that users can call that is specific only to this contract.  We don't use the parameter here so we should just remove it.

ZenGround0: I see this came from #95.  It doesn't make that much sense in practice.  The argument is that payee or payer are smart contracts that can't be upgraded.  I think this is more future proofing than we need realistically.

aarshkshah1992: Okay have addressed this.

PR #131, opened 2025-08-11
Implement Key-Value Metadata Structure (after renaming)

pali101:
## Summary

This PR completes the key-value metadata work introduced in #79 but includes changes made after rename PR.

Core Change: Switch metadata from a single string to a flexible key-value model.

Please see: #79 for more details.


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -105,8 +125,9 @@ contract FilecoinWarmStorageService is
 
     // Decode structure for data set creation extra data
     struct DataSetCreateData {
-        string metadata;
         address payer;
+        string[] metadataKeys;
+        bytes[] metadataValues;
```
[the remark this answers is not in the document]
rvagg: no, we don't need it anymore

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -489,6 +544,11 @@ contract FilecoinWarmStorageService is
         emit DataSetRailsCreated(
             dataSetId, pdpRailId, cacheMissRailId, cdnRailId, createData.payer, creator, createData.withCDN
```
rvagg, line 545: withCDN is in here -- but also, this seems like a duplicate event to `DataSetCreatedWithMetadata` and maybe we should just have one `DataSetCreated` event that includes all of this information in one go

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -539,19 +599,82 @@ contract FilecoinWarmStorageService is
...
+        require(
+            metadataKeys.length == metadataValues.length,
+            Errors.MetadataKeyAndValueLengthMismatch(metadataKeys.length, metadataValues.length)
+        );
 
         // Verify the signature
         verifyAddPiecesSignature(payer, info.clientDataSetId, pieceData, firstAdded, signature);
 
+        require(metadataKeys.length > 0, Errors.EmptyMetadataKeys(dataSetId));
+        require(metadataValues.length > 0, Errors.EmptyMetadataValues(dataSetId));
```
rvagg, line 615: same problem, this should be ok, these error types can be deleted too

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -539,19 +599,82 @@ contract FilecoinWarmStorageService is
...
+                    Errors.MetadataKeyExceedsMaxLength(k, MAX_KEY_LENGTH, bytes(key).length)
+                );
+                require(
+                    value.length <= MAX_VALUE_LENGTH,
+                    Errors.MetadataValueExceedsMaxLength(k, MAX_VALUE_LENGTH, value.length)
+                );
+                dataSetPieceMetadata[dataSetPieceId][key] = value;
+                dataSetPieceMetadataKeys[dataSetPieceId].push(key);
+            }
+            emit PieceMetadataAdded(dataSetPieceId, metadataKeys, metadataValues);
```
rvagg, line 646: I know this event hasn't changed name but this really is just a `PieceAdded` event, maybe we should go with that.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -968,13 +1097,111 @@ contract FilecoinWarmStorageService is
...
+    }
+
+    /**
+     * @notice Get the metadata for a specific piece and key by dataSetId and pieceId
+     * @param dataSetId The ID of the data set
+     * @param pieceId The ID of the piece within the data set
+     * @param key The metadata key (as string) to look up (e.g., "filename", "contentType", etc.)
+     * @return The metadata value as bytes
+     */
+    function getPieceMetadataByIds(uint256 dataSetId, uint256 pieceId, string calldata key)
```
rvagg, line 1136: The fact that you have a ByIds and one with the composite suggests we might not have the right abstraction, we should only need one of those.

I think for now we go with the dataSetId and pieceId, and you introduce the composite as a separate PR to discuss differently.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1307,7 +1534,6 @@ contract FilecoinWarmStorageService is
                 payer: storageInfo.payer,
                 payee: storageInfo.payee,
                 commissionBps: storageInfo.commissionBps,
-                metadata: storageInfo.metadata,
                 pieceMetadata: storageInfo.pieceMetadata,
```
rvagg, line 1537: needs to go

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -968,13 +1097,111 @@ contract FilecoinWarmStorageService is
...
+        return dataSetPieceMetadataKeys[dataSetPieceId];
+    }
+
+    /**
+     * @notice Get all metadata keys and values for a specific data set piece
+     * @param dataSetPieceId The composite ID for (dataSetId, pieceId)
+     * @return keys An array of metadata keys corresponding to the data set piece
+     * @return values An array of metadata values corresponding to the keys
+     */
+    function getPieceMetadataAllKeys(uint256 dataSetPieceId)
```
rvagg, line 1171: I think we might have too many accessors for this and need to economise; I think we could trim down to the "get me all keys/values". I can see a case for "get me this key" but I'm not sure it's strong enough to add in this first pass, how about we add it later when we need to. For now, just one each for data set and piece that returns (keys[], values[])

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -875,10 +956,82 @@ contract FilecoinWarmStorageService is
...
+        (address payer, string[] memory keys, bytes[] memory values, bytes memory signature) =
+            abi.decode(extraData, (address, string[], bytes[], bytes));
+
+        return DataSetCreateData({payer: payer, metadataKeys: keys, metadataValues: values, signature: signature});
+    }
 
-        return DataSetCreateData({metadata: metadata, payer: payer, withCDN: withCDN, signature: signature});
+    function checkMetadataWithCDNEnabled(uint256 dataSetId) internal view returns (bool) {
+        string[] memory keys = dataSetMetadataKeys[dataSetId];
+        bytes32 keyHash = keccak256(abi.encodePacked("withCDN"));
```
rvagg, line 967: we weren't going to do abi encoded keys in this thing, that's what the length limitation was to deal with

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -69,6 +70,15 @@ contract FilecoinWarmStorageService is
     uint256 private constant TIB_IN_BYTES = GIB_IN_BYTES * 1024; // 1 TiB in bytes
     uint256 private constant EPOCHS_PER_MONTH = 2880 * 30;
 
+    // ID bits for composite IDs
+    uint256 private constant ID_BITS = 128;
```
rvagg, line 74: not used in here

juliangruber (changes requested):
Change lgtm but I have the same questions as @rvagg, primarily
- why is `metadataValues` of type `bytes`?
- can we include metadata values in the create data set signature?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -97,8 +107,17 @@ contract FilecoinWarmStorageService is
...
-    mapping(uint256 => mapping(uint256 => string)) private dataSetPieceMetadata;
+
+    // dataSetId => PieceId => (key => value)
+    mapping(uint256 => mapping(uint256 => mapping(string => string))) public dataSetPieceMetadata;
+    // dataSetId => PieceId => array of keys
+    mapping(uint256 => mapping(uint256 => string[])) internal dataSetPieceMetadataKeys;
+
+    // Mapping from data set ID to key value pair metadata
+    // dataSetId => (key => value)
+    mapping(uint256 => mapping(string => string)) public dataSetMetadata;
```
wjmelements, line 118: implement public view in FilecoinWarmStorageServiceStateView

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -829,10 +938,30 @@ contract FilecoinWarmStorageService is
...
+    }
+
+    /**
+     * @notice Returns true if `key` exists in `metadataKeys`.
+     * @param metadataKeys The array of metadata keys
+     * @param key The metadata key to look up
+     * @return True if key exists; false otherwise.
+     */
+    function hasMetadataKey(string[] memory metadataKeys, string memory key) internal pure returns (bool) {
+        require(bytes(key).length <= MAX_KEY_LENGTH, "Key exceeds max length");
```
wjmelements, line 954: I don't think you need this size check since this is only going to check keys we defined ourselves.

also it's kind of annoying that solidity `string` doesn't have `.length`.

wjmelements:
mapping is a good change; more accessible that way

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -1019,6 +1082,948 @@ contract FilecoinWarmStorageServiceTest is Test {
 
         console.log("\n=== Test completed successfully! ===");
     }
+
+    // ==== Data Set Metadata Storage Tests ====
+    function testDataSetMetadataStorage() public {
+        // Create a data set with metadata
+        (string[] memory metadataKeys, string[] memory metadataValues) = _getSingleMetadataKV("label", "Test Metadata");
```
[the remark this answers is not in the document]
rvagg: already covered in `testDataSetMetadataKeyLengthBoundaries`

service_contracts/src/FilecoinWarmStorageService.sol
rvagg: took me a bit to figure this out, but I gather your meaning was that `public` was misplaced here
changed to `internal`; I think it probably doesn't make much difference if it's `internal` or `private` so I'll just go with `internal`

Issue #119, opened 2025-08-06
upgrade to a service provider registry


jennijuju:
When designing FS, we envisioned a **Service Registry** - a unified directory where all available services and providers are listed. This registry enables clients and service operators to easily discover and select providers and their offerings. Service providers can register their cloud services just once, providing transparent information about:
- Which services are being offered
- Who is offering them
- Where they are available or restricted
- At what cost

This central registry acts as a single source of truth, allowing client applications, dApps, and onramps to efficiently find and connect with the right service providers (effectively a "yellow pages" for service offerings). Furthermore, ecosystem tools for reputation, ranking, and scoring can leverage this registry to contextualize and evaluate listed providers.

**Current State**  
Currently, our approach embeds a service provider list directly within the Warm Storage Service contract. While this serves immediate functional needs, it leads to fragmentation—each new service contract requires providers to register again, and clients must navigate multiple listings to discover all possible offerings.

**Path to the Ideal State**  
To align with our vision, we should migrate toward a unified Service Registry. This will:
- Allow providers to register and update their information in a single place
- Eliminate duplicated efforts and ensure consistency across services
- Enable ecosystem tools (discovery, selection, scoring) to work seamlessly at scale
- Make it easier for new dApps and clients to access the full breadth of available services

**Next Steps**  
I propose we begin transitioning to this registry model, steadily decoupling provider lists from individual service contracts. This will move us toward an ecosystem where registering a service once suffices for visibility and integration with all present and future services. Regarding the registry’s content, we should start lean and allow the ecosystem to evolve requirements as a Filecoin Request for Comments (FRC). For simplicity, I also propose using a key-value object similar to _ProofSetExtradata_ ([reference](https://github.com/FilOzone/filecoin-services/issues/63#issuecomment-3002705793)).

Anyone will be able to add their service to the registry by paying a one-time listing fee (to deter spam).

The registry will be a large map, **ServiceProviders**, where each provider is assigned a unique ID upon registration.

We can start with the following structure:
```solidity

enum ProductType {
    POREP,
    PDP
    // Add more as services appears 
}

struct PoRepProduct {
    bool 32GibSector;
    bool 64GiBSector;
    uint256 maxPieceSize;
    uint256 minPieceSize;
    uint256 maxTTFB; 
    uint256 price;
    bool ipniPublisher;
    // Add more PoRep-specific fields as needed
}

struct PDPProduct {
    uint256 maxPieceSize;
    uint256 minPieceSize;
[...]

PR #135, opened 2025-08-12
feat: remove service provider registry as we're moving it to it's own contract

aarshkshah1992:
This PR does the first part of #119 i.e. removes all the code and tests for the service provider registry from this service contract. We will now build a new more extensible service registry contract  as specified in #119 and upgrade the warm storage service contract to use the registry. More PRs incoming for that.

But we can merge this standalone PR without waiting on them.

### DIFF
--- a/service_contracts/src/Errors.sol
+++ b/service_contracts/src/Errors.sol
@@ -63,10 +63,6 @@ library Errors {
     /// @param dataSetId The ID of the data set whose proving period was not initialized
     error ProvingPeriodNotInitialized(uint256 dataSetId);
 
-    /// @notice Service provider is not approved or whitelisted
-    /// @param provider The address of the unapproved service provider
-    error ServiceProviderNotApproved(address provider);
-
     /// @notice The signature is invalid (recovered signer did not match expected)
     /// @param expected The expected signer address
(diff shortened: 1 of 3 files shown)


rvagg:
We should think through the implications of removing `require(approvedProvidersMap[creator], Errors.ServiceProviderNotApproved(creator));`.

1. The new registry is intended to hold more than just WarmStorage providers. So there'll be "compatible" providers, but also incompatible ones. Should we care at this level? Does it matter if you've managed to get one to make a proof set and report it here for you?
2. This would allow a provider that's not even in the registry to be a provider for this service. Do we care about that case? Is it _caveat emptor_ in here as far as the client is concerned? If you can make a PDP deal, then you can get WarmStorage to manage it for you.

aarshkshah1992:
@rvagg The goal is to integrate this contract with the new service provider registry so it will be still be a gated list of SPs who have been approved for this contract. Happy to make that change on top of this PR so we can see what's going on and merge the whole thing e2e if that works better.

aarshkshah1992:
Note: We will be raise followup PRs so that this contract still only works with an authorised list of SPs. But the details of the SPs will live in the service registry contract. This contract will only maintain a list of IDs it wants to work with.

Issue #132, opened 2025-08-11
Add Contract Size Delta Reporting to CI


rjan90:
## Summary
Enhance the existing contract size check in our CI to not only validate size limits but also **measure and report size changes** (increases/decreases) for each PR. This will provide better visibility into how changes affect contract bytecode size.

## 🔍 Current State

Our CI currently includes a contract size check ([`.github/workflows/contract-size.yml`](https://github.com/FilOzone/filecoin-services/blob/main/.github/workflows/contract-size.yml)) that:
- ✅ Validates contracts against EIP-170 runtime size limit (24,576 bytes)
- ✅ Fails the build if any contract exceeds these limits
- ✅ Uses `forge build --sizes --json` to gather size data

**What's missing:** Size change tracking between base branch and PR changes.

## Proposed Enhancement

Add delta reporting functionality that shows:

1. **Size changes per contract** (runtime and init code)
2. **Percentage changes** for easy assessment
3. **Summary of total impact** across all contracts

Rushhaabhhh:
@rjan90 can I work on this issue ?

rjan90:
@Rushhaabhhh yeah, of course. I will assign the issue to you right away.

jennijuju:
@Rushhaabhhh are you still working on this

PR #140, opened 2025-08-13
CI: Add Solidity contract size delta reporting to CI workflow

Rushhaabhhh:
This PR updates the `contract-size.yml` GitHub Actions workflow to not only check compiled contract sizes against the EIP‑170 runtime limit (24,576 bytes) but also report size changes between the PR and the base branch (main). (#132 )

### Changes : 

- Retained existing EIP‑170 limit enforcement.
- Added build step for base branch (main) contract size data on PRs using a temporary worktree
- Implemented inline comparison logic (Bash + jq) directly in the workflow
- For each contract, the workflow now logs :
  - Current runtime size
  - Base branch runtime size
  - Delta in bytes
  - Percentage change
  - Status (Increased, Decreased, Unchanged, Limit Exceeded)
- Output is in a clean, aligned table format in workflow logs for easy review.

### DIFF
--- a/service_contracts/abi/ServiceProviderRegistry.abi.json
+++ b/service_contracts/abi/ServiceProviderRegistry.abi.json
@@ -826,6 +826,25 @@
   },
   {
     "type": "function",
+    "name": "getProviderPayee",
+    "inputs": [
+      {
+        "name": "providerId",
+        "type": "uint256",
+        "internalType": "uint256"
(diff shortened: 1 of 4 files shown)


rjan90:
Hey @Rushhaabhhh! Are you still interested in getting this across the finish line? If not, I suggest we close this PR

Issue #137, opened 2025-08-13
Check in abi files in all contract areas


ZenGround0:
PDP Verifier
Warm Storage Service
Payments 

cc @hugomrdias

rvagg:
https://github.com/FilOzone/filecoin-services/pull/141#issuecomment-3186718958

my preference would be for always up-to-date ABI files

PR #141, opened 2025-08-13
Publish ABI on release 

ZenGround0:
#137 

cc @hugomrdias


BigLep:
I agree with @rvagg that we should check these `abi.json` files in and have a CI check.  That said, lets discuss/decide/document in the issue: https://github.com/FilOzone/filecoin-services/issues/137

BigLep:
For now I'm going to close this PR given the decision to not publish ABI on release but rather have ABIs checked in and have a corresponding  CI check (see https://github.com/FilOzone/filecoin-services/issues/137 ).  Feel free to reopen if want to repurpose the PR for the change in direction.

PR #144, opened 2025-08-14
Extsload

wjmelements:
Reviewer @aarshkshah1992
This reduces the size a good amount by moving view methods out of the core contract
Smart contracts using the service can use the WarmStorageView library.
SDKs can use the TODO view contract
#### Changes
* generate FilecoinWarmStorageServiceLayout.sol from FilecoinWarmStorageService.sol
* move most viewing logic out of FilecoinWarmStorageService into library WarmStorageView
* TODO finish moving view methods over
* TODO make view contract


service_contracts/.gas-snapshot
```diff
@@ -0,0 +1,32 @@
+ExtsloadTest:test_extsload() (gas: 40763)
+ExtsloadTest:test_extsloadStruct() (gas: 61981)
```
wjmelements, line 2: I mainly added this snapshot to measure these two when I was verifying that my assembly was better than Uniswap's. I plan to remove it before merging but I'm keeping it for now to make sure there aren't significant regressions later on

wjmelements:
Planning to make the view contract autogenerate from the view library.

service_contracts/.gas-snapshot
wjmelements: 3282ce5

wjmelements:
> > Smart contracts using the service can use the WarmStorageView library.
> 
> SDKs can use the FilecoinWarmStorageServiceView contract
> 
> 
> 
> Is there a reason why one would have the Filecoin prefix and the other not?

No. I will make it consistent.

service_contracts/src/Extsload.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+pragma solidity ^0.8.20;
+
+contract Extsload {
+    function extsload(bytes32 slot) external view returns (bytes32) {
```
aarshkshah1992, line 5: @wjmelements Should these functions have no access control so only the owner/a specific contract can read these memory slots ?

wjmelements: There's no reason to restrict on-chain reads. Access control is for writes.

service_contracts/src/lib/FilecoinWarmStorageServiceStateLibrary.sol
```diff
@@ -0,0 +1,232 @@
...
+        returns (uint256)
+    {
+        return uint256(service.extsload(keccak256(abi.encode(dataSetId, PROVING_ACTIVATION_EPOCH_SLOT))));
+    }
+
+    function provingDeadlines(FilecoinWarmStorageService service, uint256 setId) public view returns (uint256) {
+        return uint256(service.extsload(keccak256(abi.encode(setId, PROVING_DEADLINES_SLOT))));
+    }
+
+    function getMaxProvingPeriod(FilecoinWarmStorageService service) public view returns (uint64) {
```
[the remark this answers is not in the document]
wjmelements: I make these public because I want to access the ABI to autogenerate the View contract. But this means that they will be accessed via `delegatecall`. This was not my intent so I plan to autogenerate another Library where all the methods are internal. This will give state viewers the option of `delegatecall` vs embedding.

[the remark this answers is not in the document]
wjmelements: created StateInternalLibrary.sol in 198f6f4

service_contracts/tools/generate_storage_layout.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+echo // SPDX-License-Identifier: Apache-2.0 OR MIT
+echo pragma solidity ^0.8.20\;
+echo
+echo // Generated with $0 $@
+echo
+
+forge inspect --json $1 storageLayout \
```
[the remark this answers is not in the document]
wjmelements: Reported to foundry https://github.com/foundry-rs/foundry/issues/11354

wjmelements:
https://github.com/filecoin-project/curio/tree/rename

rvagg:
No, Curio doesn't know about WarmStorage, only PDPVerifier, the integration work for here is in synapse-sdk `rvagg/rename` branch. But we have metadata to do on top of this and that to be integrated too.

PR #149, opened 2025-08-15
chore: add FilecoinWarmStorageService dataset metadata schema

jennijuju:
add metadata schema for downstream tooling integration 

blocked by #131


rvagg:
No, we should do something with this; it's more of a TODO than a solid proposal. IMO this ends up more like the multicodec registry - just a table with descriptions of "known metadata".

PR #160, opened 2025-08-21
fix: make deploy-warm-storage-view.sh work

rvagg:
matching the other deploy contracts, keystore's don't have to be password locked


rjan90 (approved):

rvagg:
Made a few more changes while getting this to work, commit is now:

* don't require PASSWORD (keystores can be passwordless)
* add contract name specifier
* inherit nonce, or figure it out from the chain

PR #161, opened 2025-08-21
Allow FilCDN to terminate the CDN service while keeping the storage service

pyropy:
In order to stay compliant FilCDN needs a mechanisms that would enable it to terminate service for sanctioned users. 

This pull request introduces: 
- New method `terminateCDNService` which enables only FilCDN to terminate CDN-related payment rails
- Refactor of `railTerminated` callback to support termination of CDN-related payment rails
- Separation of FilCDN addresses to `filCDNControllerAddress` and `filCDNBeneficiaryAddress`

More details about FilCDN address sanction screening and service termination is outlined in [this proposal](https://www.notion.so/spacemeridian/Sanction-screening-of-FilCDN-clients-256cdd5cccdb80c9be5efa3a42d6174f).

Closes https://github.com/filcdn/roadmap/issues/72

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -1521,7 +1521,7 @@ contract FilecoinWarmStorageService is
         if (info.pdpEndEpoch == 0 && railId == info.pdpRailId) {
             info.pdpEndEpoch = endEpoch;
             emit PDPPaymentTerminated(dataSetId, endEpoch, info.pdpRailId);
-        } else if (info.cdnEndEpoch == 0 && (railId == info.cdnRailId || railId == info.cacheMissRailId)) {
+        } else if (info.cdnEndEpoch == 0 && (railId == info.cacheMissRailId || railId == info.cdnRailId)) {
             info.cdnEndEpoch = endEpoch;
             emit CDNPaymentTerminated(dataSetId, endEpoch, info.cacheMissRailId, info.cdnRailId);
         }


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -640,6 +644,30 @@ contract FilecoinWarmStorageService is
...
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check if already terminated
+        require(info.paymentEndEpoch == 0, Errors.DataSetPaymentAlreadyTerminated(dataSetId));
+
+        // Check authorization
+        require(msg.sender == filCDNAddress, Errors.OnlyFilecoinCDNAllowed(filCDNAddress, msg.sender));
+
+        Payments payments = Payments(paymentsContractAddress);
+        payments.terminateRail(info.cacheMissRailId);
```
pyropy, line 660: It would be useful if there was a mechanism to terminate the rail without settlement. That would prevent us from recieving funds from sanctioned address.

service_contracts/src/Errors.sol
```diff
@@ -176,4 +176,13 @@ library Errors {
     /// @notice Data set does not exist for the given rail
     /// @param railId The rail ID
     error DataSetNotFoundForRail(uint256 railId);
+
+    /// @notice FilecoinCDN service is not configured for the given data set
+    /// @param dataSetId The data set ID
+    error FilecoinCDNServiceNotConfigured(uint256 dataSetId);
```
juliangruber, line 182:
```suggestion
    /// @notice CDN service is not configured for the given data set
    /// @param dataSetId The data set ID
    error CDNServiceNotConfigured(uint256 dataSetId);
```

I think so far we've been calling it just "CDN" in this contract, or the correcnt name "FilCDN"

```suggestion
    /// @notice FilCDN service is not configured for the given data set
    /// @param dataSetId The data set ID
    error FilCDNServiceNotConfigured(uint256 dataSetId);
```

service_contracts/src/Errors.sol
```diff
@@ -176,4 +176,13 @@ library Errors {
...
+
+    /// @notice Only the FilecoinCDN address can call this function
+    /// @param expected The expected FilecoinCDN address
+    /// @param actual The caller address
+    error OnlyFilecoinCDNAllowed(address expected, address actual);
```
juliangruber, line 187:
```suggestion
    error OnlyCDNAllowed(address expected, address actual);
```

service_contracts/src/Errors.sol
```diff
@@ -176,4 +176,13 @@ library Errors {
     /// @notice Data set does not exist for the given rail
     /// @param railId The rail ID
     error DataSetNotFoundForRail(uint256 railId);
+
+    /// @notice FilecoinCDN service is not configured for the given data set
+    /// @param dataSetId The data set ID
+    error FilecoinCDNServiceNotConfigured(uint256 dataSetId);
+
+    /// @notice Only the FilecoinCDN address can call this function
+    /// @param expected The expected FilecoinCDN address
```
juliangruber, line 185:
```suggestion
    /// @param expected The expected CDN address
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -640,6 +644,30 @@ contract FilecoinWarmStorageService is
...
+        DataSetInfo storage info = dataSetInfo[dataSetId];
+        require(info.withCDN, Errors.FilecoinCDNServiceNotConfigured(dataSetId));
+        require(info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check if already terminated
+        require(info.paymentEndEpoch == 0, Errors.DataSetPaymentAlreadyTerminated(dataSetId));
+
+        // Check authorization
+        require(msg.sender == filCDNAddress, Errors.OnlyFilecoinCDNAllowed(filCDNAddress, msg.sender));
```
juliangruber, line 657: We need to discuss the address...Since the main purpose of `filCDNAddress` it to receive funds, I want to make it a multisig (for safety). If the worker wants to be able to automatically terminate rails, that means it needs to have access to the same address.

I think it makes sense to separate the treasury from the operator. Does that make sense? If so, I think we can add a 2nd address, called `filCDNOperatorAddress`. And maybe rename the existing one to `filCDNTreasuryAddress`. cc @patrickwoodhead

service_contracts/src/FilecoinWarmStorageService.sol
juliangruber: If this function causes settlement, that's a problem and puts us in legal risk. We have to prevent that. Could we add an argument or another method for destroying without settling? Or, can we refund before terminating, so that the settlement is a no-op?

juliangruber: cc @aarshkshah1992

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -640,6 +644,30 @@ contract FilecoinWarmStorageService is
...
+        Payments payments = Payments(paymentsContractAddress);
+        payments.terminateRail(info.cacheMissRailId);
+        payments.terminateRail(info.cdnRailId);
+
+        emit CDNServiceTerminated(msg.sender, dataSetId, info.cacheMissRailId, info.cdnRailId);
+
+        // Clear all CDN-related fields in the data set info
+        info.withCDN = false;
+        info.cacheMissRailId = 0;
+        info.cdnRailId = 0;
```
juliangruber, line 668: `terminateService` doesn't reset the rail IDs. I wonder why that is? @aarshkshah1992

juliangruber (changes requested):

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -181,7 +186,8 @@ contract FilecoinWarmStorageService is
         address _pdpVerifierAddress,
         address _paymentsContractAddress,
         address _usdfcTokenAddress,
-        address _filCDNAddress
+        address _filCDNAddress,
```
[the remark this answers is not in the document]
aarshkshah1992: It's use has been replaced by `_filCDNBeneficiaryAddress `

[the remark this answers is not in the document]
pyropy: > We don't need the _filCDNAddress  in the constructor anymore, right ?
> It's use has been replaced by _filCDNBeneficiaryAddress 

It is actually used but for different things.  Take a look at https://github.com/FilOzone/filecoin-services/blob/e720354c178e1816ae2be18064885e04defdccc4/service_contracts/src/FilecoinWarmStorageService.sol#L660-L661

We've split the responsibilities of FilCDN addresses into two parts: controller and beneficiary. Controller address is used for operations like terminating the deals while beneficiary address should be a multisig used as a treasury.

service_contracts/src/FilecoinWarmStorageService.sol
aarshkshah1992: I mean the payments contract does it internally after the rail is finalised. See `finalisedRail` in the payments contract. Over here, setting a non-zero end epoch on the dataset tells us the rail is terminated.

service_contracts/src/FilecoinWarmStorageService.sol
aarshkshah1992: @pyropy This method does not settle anything. Simply marks an `endEpoch` on the rail. Please take a look at the implementation of `terminateRail` on the payments contract and let me know if you still have concerns.

pyropy: @aarshkshah1992 I thought that calling the `terminateRail` will trigger `settleAccountLockupBeforeAndAfterForRail` which settles the rail before and after the method call?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -640,6 +648,30 @@ contract FilecoinWarmStorageService is
...
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check if already terminated
+        require(info.paymentEndEpoch == 0, Errors.DataSetPaymentAlreadyTerminated(dataSetId));
+
+        // Check authorization
+        require(msg.sender == filCDNControllerAddress, Errors.OnlyCDNAllowed(filCDNControllerAddress, msg.sender));
+
+        Payments payments = Payments(paymentsContractAddress);
+        payments.terminateRail(info.cacheMissRailId);
```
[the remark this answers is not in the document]
pyropy: Refactored in https://github.com/FilOzone/filecoin-services/pull/161/commits/ddec5582145417528269b753ebf6be22d97ad935

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1300,9 +1363,12 @@ contract FilecoinWarmStorageService is
         uint256 dataSetId = railToDataSet[railId];
         require(dataSetId != 0, Errors.DataSetNotFoundForRail(railId));
         DataSetInfo storage info = dataSetInfo[dataSetId];
-        if (info.paymentEndEpoch == 0) {
+        if (info.paymentEndEpoch == 0 && info.pdpRailId == railId) {
             info.paymentEndEpoch = endEpoch;
             emit PaymentTerminated(dataSetId, endEpoch, info.pdpRailId, info.cacheMissRailId, info.cdnRailId);
+        } else if (info.cdnEndEpoch == 0 && (railId == info.cdnRailId || railId == info.cacheMissRailId)) {
+            info.cdnEndEpoch = endEpoch;
```
pyropy, line 1370: @aarshkshah1992 Since we're going to get `railTerminated` callback twice for termination of payment rails should we set largest of the two `endEpochs`?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -788,6 +799,34 @@ contract FilecoinWarmStorageService is
...
+        );
+        require(keccak256(bytes(withCDN)) == keccak256("true"), Errors.CDNServiceNotConfigured(dataSetId));
+        require(info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check if already terminated
+        require(info.cdnEndEpoch == 0, Errors.CDNPaymentAlreadyTerminated(dataSetId));
+
+        // Check authorization
+        require(msg.sender == filCDNControllerAddress, Errors.OnlyCDNAllowed(filCDNControllerAddress, msg.sender));
```
jennijuju, line 817: can the msg.sender be the client?

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
juliangruber: We should rename it to `filCDNControllAddress`

service_contracts/src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
```diff
@@ -101,6 +101,7 @@ library FilecoinWarmStorageServiceStateInternalLibrary {
         info.commissionBps = uint256(info8[5]);
         info.clientDataSetId = uint256(info8[6]);
         info.paymentEndEpoch = uint256(info8[7]);
+        info.cdnEndEpoch = uint256(info8[8]);
```
juliangruber, line 104: this needs a subgraph update

service_contracts/src/Errors.sol
```diff
@@ -13,14 +13,16 @@ library Errors {
...
+        FilecoinCDNController,
         /// Creator address
         Creator,
         /// Payer address
         Payer,
         /// Service provider address
-        ServiceProvider
+        ServiceProvider,
+        /// FilecoinCDN treasury address
+        FilecoinCDNTreasury
```
juliangruber, line 25:
```suggestion
        /// FilCDN beneficiary address
        FilCDNBeneficiary
```

service_contracts/src/Errors.sol
```diff
@@ -13,14 +13,16 @@ library Errors {
         Payments,
         /// USDFC contract address
         USDFC,
-        /// FilecoinCDN service contract address
-        FilecoinCDN,
+        /// FilecoinCDN controller address
+        FilecoinCDNController,
```
juliangruber, line 17:
```suggestion
        /// FilCDN controller address
        FilCDNController,
```

service_contracts/src/Errors.sol
```diff
@@ -177,6 +179,19 @@ library Errors {
     /// @param railId The rail ID
     error DataSetNotFoundForRail(uint256 railId);
 
+    /// @notice CDN service is not configured for the given data set
+    /// @param dataSetId The data set ID
+    error CDNServiceNotConfigured(uint256 dataSetId);
+
+    /// @notice Only the FilecoinCDN address can call this function
```
juliangruber, line 186:
```suggestion
    /// @notice Only the FilCDN controller can call this function
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -63,10 +63,16 @@ contract FilecoinWarmStorageService is
...
 
+    event CDNServiceTerminated(
+        address indexed caller, uint256 indexed dataSetId, uint256 cacheMissRailId, uint256 cdnRailId
+    );
+
     event PaymentTerminated(
         uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
     );
 
+    event CDNPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 cacheMissRailId, uint256 cdnRailId);
```
juliangruber, line 74: It's confusing to have `PaymentTerminated` and `CDNPaymentTerminated`. Will `CDNPaymentTerminated` also be emitted when all payments have been terminated?

I understand the first event is emitted when everything related to a data set is terminated, the second only when the CDN part is terminated. 

Any suggestions?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -135,6 +142,7 @@ contract FilecoinWarmStorageService is
         uint256 commissionBps; // Commission rate for this data set (dynamic based on whether the client purchases CDN add-on)
         uint256 clientDataSetId; // ClientDataSetID
         uint256 paymentEndEpoch; // 0 if payment is not terminated
+        uint256 cdnEndEpoch; // 0 if CDN is not terminated
```
juliangruber, line 145: this is also confusing, what's the difference between terminating the payment and the CDN (I'm asking from a user's perspective)?

What about we split this up and have `pdpPaymentTerminated` and `cdnPaymentTerminated`, and when the data set is terminated, both events will get emitted?

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: Renamed `PaymentTerminated` to `PDPPaymentTerminated` and removed CDN-specific rail ids from the event https://github.com/FilOzone/filecoin-services/pull/161/commits/3ed26569d371873159f5b017162c71904d3f2544

bajtos:
I want to flag that we need to update the deployment scripts and replace `$FILCDN_ADDRESS` with two new configuration flags.

See e.g. here:
https://github.com/FilOzone/filecoin-services/blob/5a322b2b21138112c56563267985be7da182da12/service_contracts/tools/deploy-warm-storage-calibnet.sh#L30

I don't have a strong opinion on whether such changes must be done as part of this pull request or whether it can be a follow-up PR.

service_contracts/src/FilecoinWarmStorageService.sol
aarshkshah1992: @pyropy No, set the smaller of the two epochs as that is the real epoch of rail termination.

Actually wait, let's only terminate the storage rail in `terminateService` if the CDN rails have already been terminated.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -66,10 +66,14 @@ contract FilecoinWarmStorageService is
         address indexed caller, uint256 indexed dataSetId, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
     );
 
-    event PaymentTerminated(
-        uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
+    event CDNServiceTerminated(
+        address indexed caller, uint256 indexed dataSetId, uint256 cacheMissRailId, uint256 cdnRailId
```
aarshkshah1992, line 70: The caller will always be the CDN controller right ?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -66,10 +66,14 @@ contract FilecoinWarmStorageService is
...
     );
 
+    event PDPPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId);
+
+    event CDNPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 cacheMissRailId, uint256 cdnRailId);
```
aarshkshah1992, line 75: What's the difference between `CDNPaymentTerminated ` and `CDNServiceTerminated ` ?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -233,27 +239,30 @@ contract FilecoinWarmStorageService is
...
-        require(_filCDNAddress != address(0), "Filecoin CDN address cannot be zero");
-        filCDNAddress = _filCDNAddress;
+        require(_filCDNControllerAddress != address(0), "Filecoin CDN address cannot be zero");
+        filCDNControllerAddress = _filCDNControllerAddress;
 
         require(_pdpVerifierAddress != address(0), Errors.ZeroAddress(Errors.AddressField.PDPVerifier));
         require(_paymentsContractAddress != address(0), Errors.ZeroAddress(Errors.AddressField.Payments));
         require(_usdfcTokenAddress != address(0), Errors.ZeroAddress(Errors.AddressField.USDFC));
-        require(_filCDNAddress != address(0), Errors.ZeroAddress(Errors.AddressField.FilecoinCDN));
+        require(_filCDNControllerAddress != address(0), Errors.ZeroAddress(Errors.AddressField.FilCDNController));
```
aarshkshah1992, line 257: already checked for this on line 251. maybe use this error code in line 251 ?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -860,6 +869,34 @@ contract FilecoinWarmStorageService is
...
+        );
+        require(keccak256(bytes(withCDN)) == keccak256("true"), Errors.CDNServiceNotConfigured(dataSetId));
+        require(info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check if already terminated
+        require(info.cdnEndEpoch == 0, Errors.CDNPaymentAlreadyTerminated(dataSetId));
+
+        // Check authorization
+        require(msg.sender == filCDNControllerAddress, Errors.OnlyCDNAllowed(filCDNControllerAddress, msg.sender));
```
aarshkshah1992, line 887: make this a modifier and put that on the function

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1077,6 +1114,30 @@ contract FilecoinWarmStorageService is
         return false;
     }
 
+    /**
+     * @notice Deletes `key` if it exists in `metadataKeys`.
+     * @param key The array of metadata keys
+     * @return Modified array of metadata keys
+     */
+    function deleteMetadataKey(string[] memory metadataKeys, string memory key)
```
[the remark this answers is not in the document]
juliangruber: In https://github.com/FilOzone/filecoin-services/blob/982b5a6e32b800322e97fc1c92bb599ac02b4300/service_contracts/src/FilecoinWarmStorageService.sol#L448 there's an invariant that the keys length should equal the values length. That won't be true any more after this function. I think we need to remove the metadata value at the same index.

And if the index isn't the last element of the array, we need to reshuffle it so as to not have any holes in it (see eg https://ethereum.stackexchange.com/questions/1527/how-to-delete-an-element-at-a-certain-index-in-an-array))

juliangruber (changes requested):

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: This must've slipped during the rebase. Fixed.

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
pyropy: > can we add some tests specifically for this function ?

I'm hesitant to make this function public. But, we are now testing metadata keys & value length after the CDN termination.

```solidity
(metadataKeys, metadataValues) = viewContract.getAllDataSetMetadata(dataSetId);
assertTrue(metadataKeys.length == 0, "Metadata keys should be empty after termination");
assertTrue(metadataValues.length == 0, "Metadata values should be empty after termination");
```

@juliangruber Thanks for pointing this out! I've modified the function to delete the reduce the size of an array after the deletion.

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: > Actually wait, let's only terminate the storage rail in terminateService if the CDN rails have already been terminated.

Could you explain the motivation behind that? I'm not sure I understand why should we do that.

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: Not necessarily as in the future we may allow payer to call this function too.

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
juliangruber: The function is still not removing values from `metadataValues`

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -860,6 +869,34 @@ contract FilecoinWarmStorageService is
         emit ServiceTerminated(msg.sender, dataSetId, info.pdpRailId, info.cacheMissRailId, info.cdnRailId);
     }
 
+    function terminateCDNService(uint256 dataSetId) external {
```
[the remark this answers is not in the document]
pyropy: @aarshkshah1992 Rates for PDP rail and CDN-related rails should be calculated and managed separately. In case data set had `withCDN` during creation but was terminated rate for PDP payment rail should remain unchanged. 

Once the CDN service is disabled for the `withCDN` should be removed from the metadata meaning that this section inside `updatePaymentRates` won't be triggered anymore:

```solidity
        // Update the CDN rail payment rates, if applicable
        if (hasMetadataKey(dataSetMetadataKeys[dataSetId], METADATA_KEY_WITH_CDN)) {
            (uint256 newCacheMissRatePerEpoch, uint256 newCDNRatePerEpoch) = _calculateCDNRates(totalBytes);

            uint256 cacheMissRailId = dataSetInfo[dataSetId].cacheMissRailId;
            payments.modifyRailPayment(cacheMissRailId, newCacheMissRatePerEpoch, 0);
            emit RailRateUpdated(dataSetId, cacheMissRailId, newCacheMissRatePerEpoch);

            uint256 cdnRailId = dataSetInfo[dataSetId].cdnRailId;
            payments.modifyRailPayment(cdnRailId, newCDNRatePerEpoch, 0);
            emit RailRateUpdated(dataSetId, cdnRailId, newCDNRatePerEpoch);
        }
```

service_contracts/src/FilecoinWarmStorageService.sol
aarshkshah1992: @pyropy I meant that we should NOT terminate CDN rails in `terminateService ` if they have already been terminated but looks like that's been taken care of

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
aarshkshah1992: @pyropy Sure, but please can we have a test for this as described above ?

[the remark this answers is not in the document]
pyropy: @aarshkshah1992 Added in https://github.com/FilOzone/filecoin-services/pull/161/commits/4df48b728d41eb06f7e3f12aad2abb82f5e65c2f

service_contracts/src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
pyropy: I have opened https://github.com/FilOzone/filecoin-services/pull/170 but the current subgraph is outdated and I'm waiting for a new subgraph before updating it.

aarshkshah1992 (approved):
lgtm but need @juliangruber to review and approve as well.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -866,6 +884,30 @@ contract FilecoinWarmStorageService is
...
 
+    function terminateCDNService(uint256 dataSetId) external onlyFilCDNController {
+        require(
+            hasMetadataKey(dataSetMetadataKeys[dataSetId], METADATA_KEY_WITH_CDN),
+            Errors.FilCDNServiceNotConfigured(dataSetId)
+        );
+
+        DataSetInfo storage info = dataSetInfo[dataSetId];
+        require(info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
```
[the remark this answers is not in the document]
pyropy: If the CDN service hasn’t been enabled, the withCDN flag will be absent from the metadata, and an error will already have been thrown before these checks are reached. In practice, these checks serve only as an extra safeguard and are unlikely to ever be triggered.

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -1283,6 +1288,249 @@ contract FilecoinWarmStorageServiceTest is Test {
...
+        FilecoinWarmStorageService.DataSetInfo memory info = viewContract.getDataSet(dataSetId);
+        vm.prank(pdpServiceWithPayments.filCDNControllerAddress()); // FilCDN terminates
+        vm.expectEmit(true, true, true, true);
+        emit FilecoinWarmStorageService.CDNServiceTerminated(
+            filCDNController, dataSetId, info.cacheMissRailId, info.cdnRailId
+        );
+        pdpServiceWithPayments.terminateCDNService(dataSetId);
+
+        // 5. Assertions
+        // Check if CDN data is cleared
```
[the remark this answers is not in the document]
pyropy: Added https://github.com/FilOzone/filecoin-services/pull/161/commits/c768750477ffe218ab6071d3335b536d2520a476

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -67,10 +67,14 @@ contract FilecoinWarmStorageService is
...
 
-    event PaymentTerminated(
-        uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
+    event CDNServiceTerminated(
+        address indexed caller, uint256 indexed dataSetId, uint256 cacheMissRailId, uint256 cdnRailId
     );
 
+    event PDPPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId);
+
+    event CDNPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 cacheMissRailId, uint256 cdnRailId);
```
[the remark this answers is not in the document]
pyropy: Added tests for both events  https://github.com/FilOzone/filecoin-services/pull/161/commits/3c5aa0ba6ea1159b21535203b04b52813e13fdd5

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -67,10 +67,14 @@ contract FilecoinWarmStorageService is
         address indexed caller, uint256 indexed dataSetId, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
     );
 
-    event PaymentTerminated(
-        uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
+    event CDNServiceTerminated(
+        address indexed caller, uint256 indexed dataSetId, uint256 cacheMissRailId, uint256 cdnRailId
     );
 
+    event PDPPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId);
```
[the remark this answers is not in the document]
pyropy: Added https://github.com/FilOzone/filecoin-services/pull/161/commits/3c5aa0ba6ea1159b21535203b04b52813e13fdd5

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1083,6 +1125,40 @@ contract FilecoinWarmStorageService is
...
+            if (currentKeyBytes.length == keyLength && keccak256(currentKeyBytes) == keyHash) {
+                // Shift elements left to fill the gap
+                for (uint256 j = i; j < len - 1; j++) {
+                    metadataKeys[j] = metadataKeys[j + 1];
+                }
+
+                delete metadataKeys[len - 1];
+                assembly {
+                    mstore(metadataKeys, sub(len, 1))
+                }
```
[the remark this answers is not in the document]
pyropy: Sadly, `pop` is only available for storage arrays and not for `string[] memory`. Do you think we should modify `deleteMetadataKey` to operate directly on state?

PR #181, opened 2025-08-27
Fix StateLibrary json Makefile recipe

wjmelements:
Reviewer @rvagg
There are a few problems with the `FilecoinWarmStorageServiceStateLibrary.json` recipe.
It hides important information about why the build failed.
It behaves as if the recipe succeeded even when it failed.
#### Changes
* do not hide the build errors
* do not proceed as if successful when the build fails
* no need to --force the build


rvagg (approved):
OK, but I've bumped into a circular problem multiple times of not being able to build because things are out of sync so the compiler complains, and then having to fall back to manually removing things or restoring from HEAD just to get things happy, so this was one attempt to square a circle.

PR #180, opened 2025-08-27
feat(ci): gen, abi-gen, coverage

rvagg:
Closes: https://github.com/FilOzone/filecoin-services/pull/148
Fixes: https://github.com/FilOzone/filecoin-services/issues/158
Fixes: https://github.com/FilOzone/filecoin-services/issues/137

Builds on https://github.com/FilOzone/filecoin-services/pull/173 and extends
the Makefile further to do a lot of the work.

PR #188, opened 2025-08-29
feat: semi-incremental abi generation

rvagg:
Building on discussion in https://github.com/FilOzone/filecoin-services/pull/180#discussion_r2308191369 about the current `update-abi` target.

There's a couple of compromises here over the pure pattern target approach, because:

1. If we rely on `forge inspect` we get a compile-all each time we call it, so it's just slow and tedious; we could go with that, but we have the build output anyway so we can just rely on that and get the same output plus pull in build to our dependency chain
2. I can't just do `abi/%.abi.json: out/%.sol/%.json` to rely on those because if they don't exist then they're not included in our pattern at parse time. This could work _if_ we wanted to dump all of the compiled contracts into abi/, but there's a _lot_ of contracts to dump in there. So we just have a whitelist and a macro to strictly specify that we have a rule per output file we want.

I also tried out a variant of this that put in some of our dependencies to capture their ABI at the current commit, but it was more complex so I'll leave that can of worms alone and we'll just focus on the contracts in here.

### DIFF
--- a/service_contracts/Makefile
+++ b/service_contracts/Makefile
@@ -103,6 +103,11 @@ clean:
 	forge clean
 	rm -rf out cache
 
+# Clean everything including generated files and ABIs
+.PHONY: clean-all
+clean-all: clean clean-gen clean-abi
+	@echo "All artifacts cleaned"
+
 # Format code
 .PHONY: fmt
 fmt:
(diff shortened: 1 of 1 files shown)


wjmelements (approved):

service_contracts/Makefile
```diff
@@ -132,18 +137,41 @@ contract-size-check:
...
-	@echo "ABIs updated in abi/ directory"
+	@echo "Extracting ABI for $(1)..."
+	@jq '.abi' $$< > $$@
+endef
+
+# Generate rules for each contract using the template above
+$(foreach contract,$(ABI_CONTRACTS),$(eval $(call ABI_RULE,$(contract))))
+
+# JSON files depend on build
+out/%.sol/%.json: build
```
wjmelements, line 165: I think you can't have two stems in the recipe, but can do

out/%.json

rvagg: `out/$(1).sol/$(1).json: | build` in the macro instead

wjmelements (approved):

rvagg:
OK, this now includes ServiceProviderRegistry too; rebased on main.

service_contracts/Makefile
```diff
@@ -132,18 +137,48 @@ contract-size-check:
...
+# at parse-time will be rejected by make, so we'll be explicit instead.
+define ABI_RULE
+abi/$(1).abi.json: out/$(1).sol/$(1).json | abi
+	@echo "Extracting ABI for $(1)..."
+	@jq '.abi' $$< > $$@
+
+# Mark JSON as coming from build (unless it's the library which has its own rule)
+ifneq ($(1),FilecoinWarmStorageServiceStateLibrary)
+out/$(1).sol/$(1).json: | build
+	@# Created by build
```
wjmelements, line 163: can this line be removed?

wjmelements (approved):

PR #193, opened 2025-09-02
feat: name and description in FWSS

aarshkshah1992:
We will raise a FRC down the road to standardize this so that Synapse tooling can pick up the name and description for newly deployed services. We only need to standardise the deployment event and the getters for these. 

Subsumes https://github.com/FilOzone/filecoin-services/pull/108.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -45,6 +45,7 @@ contract FilecoinWarmStorageService is
 
     // Events
     event ContractUpgraded(string version, address implementation);
+    event FilecoinServiceDeployed(string name, string description);
     event DataSetServiceProviderChanged(
         uint256 indexed dataSetId, address indexed oldServiceProvider, address indexed newServiceProvider
     );
@@ -193,6 +194,10 @@ contract FilecoinWarmStorageService is
     // directly instead of going through the view contract for more efficient gas usage.
     address public viewContractAddress;
 
+    // Service name and description (immutable after initialization)
+    string private serviceName;
+    string private serviceDescription;
+
     // Approved provider list
     mapping(uint256 => bool) internal approvedProviders;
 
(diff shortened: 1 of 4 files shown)


jennijuju (approved):

aarshkshah1992:
@rvagg Aye, the getter function should not be in this file, right ? It has to be in the library as it will only ever be used externally ?

wjmelements:
I discussed this with @jennijuju. First, we don't know about all of the metadata that we will require for each service listing, and a better way to track those would be with a key-value storage, like we do for the ServiceProviderRegistry. For example, we might want localized (eg 中文) descriptions and web interface URLs. Second, self-identification is not a great way to surface metadata, both because it can be malicious and because it can be incomplete. We resolved that name and description should not be part of the interface but instead metadata keys in a future curated service registry. I don't think we should work on such a registry until there are at least two services.

I did agree to keep the event though. I'll update this PR accordingly.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -233,6 +234,10 @@ contract FilecoinWarmStorageService is
     // The address allowed to terminate CDN services
     address private filCDNControllerAddress;
 
+    // Service information (added at the end to preserve storage layout for upgrades)
```
rvagg, line 237:
```suggestion
    // Service information
```

rvagg: Let's put these 2 right at the top of the slots so they take position 0 and 1, as per #198 - or I can do that after this merges when I rebase 198.

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -446,7 +446,9 @@ contract FilecoinWarmStorageServiceTest is Test {
             FilecoinWarmStorageService.initialize.selector,
             uint64(2880), // maxProvingPeriod
             uint256(60), // challengeWindowSize
-            filCDNController // filCDNControllerAddress
+            filCDNController, // filCDNControllerAddress
+            "Filecoin Warm Storage Service", // service name
+            "A decentralized storage service with proof-of-data-possession and payment integration" // service description
```
rvagg, line 451: 🤷 @jennijuju got anything better here?

service_contracts/src/FilecoinWarmStorageService.sol
aarshkshah1992: @rvagg Let's do it in 198 as we will need a rebase anyways.

Issue #194, opened 2025-09-02
FilecoinWarmStorageServiceStateInternalLibrary generator does not work on macOS


bajtos:
macOS 15.6, using the built-in `sed`:

```
❯ make src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
sed -e 's/public/internal/g' -e 's/StateLibrary/StateInternalLibrary/g' -e '3a\\n// Code generated - DO NOT EDIT.\n// This file is a generated binding and any changes will be lost.\n// Generated with make src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol\n' src/lib/FilecoinWarmStorageServiceStateLibrary.sol | forge fmt -r - > src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
sed: 1: "3a\\n// Code generated  ...": extra characters after \ at the end of a command
```

After I remove the offending `-e '3a\\n...'` params, the codegen works correctly.

wjmelements:
looks like https://stackoverflow.com/questions/40843994/extra-characters-after-at-the-end-of-a-command

PR #195, opened 2025-09-02
gen: Fix bsd sed newline issue

wjmelements:
Reviewer @bajtos
Fixes https://github.com/FilOzone/filecoin-services/issues/194
I spent too long trying to support both versions of sed and I don't think it's possible.
Make's newline encoding is [somehow](https://github.com/FilOzone/filecoin-services/issues/194#issuecomment-3246588792) different than bash's, even if you specify SHELL.
The easiest fix for the line insertion was to just use awk which seems to be more standardized.
#### Test plan
```sh
rm -f src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol; make src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
git diff src
```
#### Changes
* use awk
* use pattern (`%`) rule

### DIFF
--- a/service_contracts/Makefile
+++ b/service_contracts/Makefile
@@ -44,7 +44,7 @@ $(VIEW_CONTRACT): tools/generate_view_contract.sh $(LIBRARY_JSON)
 
 # Internal library generation (simple sed transform)
 %StateInternalLibrary.sol: %StateLibrary.sol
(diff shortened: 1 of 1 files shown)


rvagg (approved):
OK but I'm a bit concerned about going back to `Generated with make $@` because that was a problem we were trying to fix before with alternative shells - zsh I think was responsible for single-quotes wrapping around the target name here.

@aarshkshah1992 maybe you could try this branch out, remove the InternalLibrary and re-make it and let's see if it's unchanged?

rvagg:
Also at this point I'd be happy for this to be another bash script rather than wrestling with Make, at least then you could branch on different OSs or seds. I think `-E` on macOS might give you what you need instead of `-e`?

wjmelements:
I would love to reproduce that and fix it.

PR #199, opened 2025-09-03
fix: remove serviceName and serviceDescription properties

rvagg:
Same as https://github.com/FilOzone/filecoin-services/pull/196, which was accidentally re-applied due to miscommunication.


jennijuju:
CI is red

PR #207, opened 2025-09-06
Add emit `pieceCid` in `PieceAdded` event

juliangruber:
With this in place, FilCDN won't need to listen to any `PDPVerifier` events any more (and thereby not need its subgraph either.

I understand this is too late for the m2.5 calibration deployment, but hope to upgrade to this later.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -66,7 +66,7 @@ contract FilecoinWarmStorageService is
     );
     event RailRateUpdated(uint256 indexed dataSetId, uint256 railId, uint256 newRate);
     event PieceAdded(
-        uint256 indexed dataSetId, uint256 indexed pieceId, string[] keys, string[] values, Cids.Cid pieceCid
+        uint256 indexed dataSetId, uint256 indexed pieceId, Cids.Cid pieceCid, string[] keys, string[] values
     );
 
     event ServiceTerminated(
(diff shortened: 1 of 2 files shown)


rvagg:
needs a lint fix and also `make update-abi`

This doesn't break the implementation so we could do this as an upgrade to existing contracts, but I wonder if having the CID in that position is more breaking for event (subgraph) collection than putting it last? I don't know much about how subgraphs consume the data but if we did an upgrade of existing contracts and added a new entry, is it _less_ breaking and easier to adapt to if we put it at the end rather than shuffling the properties?

juliangruber:
My understanding of subgraph argument parsing is that it would be position and type dependent, so you're right I think that adding the new argument last would help there. Otoh, the argument order would be a bit unintuitive for everyone else.

I've packaged the reorder in commit e6864bd, which can be included or reverted easily.

rvagg:
@juliangruber I don't mind either way; it's _nicer_ for it to be before the metadata, but it means we're breaking data. BUT if we don't care about the data from this event between last week when we deployed the contract to when we update it then we can go ahead and break it. Is it important that we collect all of it? Or perhaps without `pieceCid` it's useless anyway? Doesn't bother me, but I don't depend on this data!

juliangruber:
> BUT if we don't care about the data from this event between last week when we deployed the contract to when we update it then we can go ahead and break it.

The subgraphs aren't live yet and can be redeployed at any time anyway, there is no impact as long as we ship a contract upgrade before we publish the subgraphs.

juliangruber:
So shall I revert and then land?

aarshkshah1992 (approved):

aarshkshah1992:
@rvagg lgtm to you ?

PR #209, opened 2025-09-08
feat: service registry: add providerId to ServiceInfo

aarshkshah1992:
Add providerId to ServiceInfo for easy provider lookup.

### DIFF
--- a/service_contracts/src/ServiceProviderRegistryStorage.sol
+++ b/service_contracts/src/ServiceProviderRegistryStorage.sol
@@ -18,6 +18,7 @@ contract ServiceProviderRegistryStorage {
 
     /// @notice Main provider information
     struct ServiceProviderInfo {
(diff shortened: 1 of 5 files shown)


hugomrdias (approved):

rvagg:
https://github.com/FilOzone/synapse-sdk/issues/208

service_contracts/src/ServiceProviderRegistryStorage.sol
```diff
@@ -18,6 +18,7 @@ contract ServiceProviderRegistryStorage {
 
     /// @notice Main provider information
     struct ServiceProviderInfo {
+        uint256 providerId; // Unique identifier for the provider
```
wjmelements, line 21: this does not belong in the struct itself. It is the key! We already have it whenever we lookup this struct!

PR #230, opened 2025-09-19
fix: FilCDN Controller & Beneficiary addresses

bajtos:
Fix the default addresses to match the latest FilCDN infra.


rjan90 (approved):
Looks correct to me/similar to the addresses posted in Slack here: https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1756826108640579

PR #233, opened 2025-09-24
feat(subgraph): add PDP offering decoding

silent-cipher:
decode provider product data and store it as a json string

### DIFF
--- a/subgraph/src/utils/contract-calls.ts
+++ b/subgraph/src/utils/contract-calls.ts
@@ -1,6 +1,6 @@
 import { Address, BigInt, Bytes, log } from "@graphprotocol/graph-ts";
 import { ServiceProviderRegistry } from "../../generated/ServiceProviderRegistry/ServiceProviderRegistry";
-import { ServiceProviderInfo } from "./types";
+import { PDPOffering, ServiceProviderInfo } from "./types";
 import { PDPVerifier } from "../../generated/PDPVerifier/PDPVerifier";
 
 export function getServiceProviderInfo(registryAddress: Address, providerId: BigInt): ServiceProviderInfo {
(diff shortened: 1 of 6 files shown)


subgraph/src/utils/contract-calls.ts
```diff
@@ -50,3 +50,26 @@ export function getPieceCidData(verifierAddress: Address, setId: BigInt, pieceId
...
+
+export function decodePDPOfferingData(registryAddress: Address, data: Bytes): PDPOffering {
+  const serviceProviderRegistryInstance = ServiceProviderRegistry.bind(registryAddress);
+
+  const pdpOfferingTry = serviceProviderRegistryInstance.try_decodePDPOffering(data);
+
+  if (pdpOfferingTry.reverted) {
+    log.warning("decodePDPOfferingData: contract call reverted for data: {}", [data.toHexString()]);
+    return PDPOffering.empty();
+  }
```
rvagg, line 62: I've been considering just removing the decode* and encode* functions off the service registry, at least as public. They just do abi.encode and abi.decode and there's no good reason to invoke the EVM just do to that. Do we have good abi.decode tools available to us at this point that we could just call to get `pdpOffering`?

silent-cipher: If subgraph is the one that blocks removing these decode* functions, then feel free to remove these
And, I think these invocations will only be limited to ( productAdded and productUpdated ) from subgraph side

silent-cipher:
removed abi duplication in subgraph

rvagg (approved):
@silent-cipher before you merge this could you add a note in subgraph/templates/subgraph.template.yaml and subgraph/src/filecoin-warm-storage-service-legacy.ts the purpose of these "legacy" files and make it clear what the path to removal of them is, just so when we have an upgrade this doesn't get lost if it falls in someone else's lap. I'm sure you can keep this in your head, but it's going to look confusing to anyone else casually browsing, and it's the kind of thing that could be left in place accidentally because the next person doesn't understand why it's there.

rvagg:
lgtm, thanks

PR #237, opened 2025-09-25
Modify CDN payment rails and add methods for usage-based payments

pyropy:
### Motivation

To support FilBeam's usage-based payment model, this PR adds methods to the FWSS contract for settling CDN payment rails without payment validators

### Description

Introduces two new methods:
- `settleCDNPaymentRails` - for settling CDN payment rails
- `topUpCDNPaymentRails` - for topping up balance in CDN payment rails (increasing fixed lockup)

Adding these two methods would enable usage-based payments for CDN-related payment rails as FilBeam contract would be able to calculate settlement amounts based on the reported egress usage and settle payment rails via `settleCDNPaymentRails`.

Validator has been removed from CDN-related payment rails as settling is done by FilBeam contract.

Users are be able to increase fixed lockup for their CDN-related payment rails by calling `topUpCDNPaymentRails` in return increasing their data set egress limit. Initial fixed lockup for CDN payment rails is set to `0` as we're unable to pass lockup amounts via metadata without increasing contract byte size due to addition of helper new functions.

### Changelog

- Remove validator from CDN payment rails
- Add methods for FilBeam operator contract to settle cdn payment rails
- Add method for increasing fixed lockup on cdn payment rails `topUpCDNPaymentRails`
- Cleanup variables, errors and events related to cdn service termination
- Remove handling of cdn payment rail termination from `railTerminated` method

Related to:
- https://github.com/filbeam/roadmap/issues/54
- https://github.com/filbeam/contracts/pull/1


service_contracts/abi/FilecoinWarmStorageService.abi.json
```diff
@@ -872,6 +918,31 @@
...
+    "name": "CDNPaymentRailsToppedUp",
+    "inputs": [
+      {
+        "name": "dataSetId",
+        "type": "uint256",
+        "indexed": true,
+        "internalType": "uint256"
+      },
+      {
+        "name": "cdnAmount",
```
bajtos, line 932: Is this the delta or the new value?

When I have 0.1 USDFS remaining in my lockup, and I top up 1 USDFC, will this event emit `1` or `1.1`?

Let's find a name that communicates this difference. For example, `totalCdnAmount` vs `cdnAmountAdded`.

The same comment applies to `cacheMissAmount` below.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -572,28 +572,32 @@ contract FilecoinWarmStorageService is
...
                 payee, // payee address from registry
-                address(this), // this contract acts as the arbiter
+                address(0), // no validator
                 0, // no service commission
-                address(this)
+                address(this) // controller
             );
             info.cacheMissRailId = cacheMissRailId;
             railToDataSet[cacheMissRailId] = dataSetId;
+            // Set lockup with default value of 0
```
bajtos, line 586: Let's discuss the initial egress allowance after a new data set is created ([Slack Discussion](https://space-meridian.slack.com/archives/C08TVNKJV7C/p1759312674268629?thread_ts=1759271890.543729&cid=C08TVNKJV7C)).

Since it's not easy to let the client specify the initial `cdnAmount` and `cacheMissAmount`, I propose to hard-code the following values for now:

- 0.7 USDFC for CDN
- 0.3 USDFC for cache-miss

@juliangruber do you have an opinion on what would be good default amounts?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,18 +976,90 @@ contract FilecoinWarmStorageService is
...
+     */
+    function settleCDNPaymentRails(uint256 dataSetId, uint256 cdnAmount, uint256 cacheMissAmount)
+        external
+        onlyFilBeamController
+    {
+        DataSetInfo storage info = dataSetInfo[dataSetId];
+        require(info.pdpRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check if CDN rails are configured (presence of rails indicates CDN was set up)
+        require(info.cdnRailId != 0 && info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
```
bajtos, line 994: Can we tell the difference between "CDN rails not configured" (withCDN was not in the metadata) and "CDN rails were terminated" (terminated by FilBeam after detecting that the client address is sanctioned)? Should we emit different errors for those two cases?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,18 +976,90 @@ contract FilecoinWarmStorageService is
...
+     */
+    function topUpCDNPaymentRails(uint256 dataSetId, uint256 cdnAmount, uint256 cacheMissAmount) external {
         DataSetInfo storage info = dataSetInfo[dataSetId];
-        require(info.cdnEndEpoch == 0, Errors.FilBeamPaymentAlreadyTerminated(dataSetId));
+        require(info.pdpRailId != 0, Errors.InvalidDataSetId(dataSetId));
 
+        // Check authorization - only payer can top up
+        require(
+            msg.sender == info.payer,
+            Errors.CallerNotPayerOrPayee(dataSetId, info.payer, info.serviceProvider, msg.sender)
```
bajtos, line 1020: The error `CallerNotPayerOrPayee` does not look correct to me. It gives the impression that the payer can top up the rails as well.

Let's find a better error. (Or create a new one - `Errors.CallerNotPayer`.)

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,18 +976,90 @@ contract FilecoinWarmStorageService is
...
+
+        // Check if CDN service is configured
+        require(
+            hasMetadataKey(dataSetMetadataKeys[dataSetId], METADATA_KEY_WITH_CDN),
+            Errors.FilBeamServiceNotConfigured(dataSetId)
+        );
+
+        // Check if cache miss and CDN rails are configured
+        require(info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
```
bajtos, line 1031: Under which conditions the rails won't be configured? 

Should we emit `Errors.FilBeamPaymentAlreadyTerminated` instead?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,18 +976,90 @@ contract FilecoinWarmStorageService is
...
+
+        Payments payments = Payments(paymentsContractAddress);
+
+        if (cacheMissAmount > 0) {
+            // Get current lockup and increment by the passed amount
+            Payments.RailView memory cacheMissRail = payments.getRail(info.cacheMissRailId);
+            payments.modifyRailLockup(
+                info.cacheMissRailId, DEFAULT_LOCKUP_PERIOD, cacheMissRail.lockupFixed + cacheMissAmount
```
bajtos, line 1039: I vaguely remember that we were discussing the following two options for implementing the top-ups:

1. "Add $X USDFC to my balance" (`new = current + amount`
2. "Increase my total balance to $X USDFC" (`assert(amount>current); new = amount`)

I don't remember the outcome of that discussion. Do you recall @pyropy if we captured the decision in writing and where?

Regardless of which approach we choose, it would be great to improve the parameter names (`cacheMissAmount`, `cdnAmount`) to clearly communicate the algorithm to the user.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1621,9 +1691,6 @@ contract FilecoinWarmStorageService is
         if (info.pdpEndEpoch == 0 && railId == info.pdpRailId) {
             info.pdpEndEpoch = endEpoch;
             emit PDPPaymentTerminated(dataSetId, endEpoch, info.pdpRailId);
-        } else if (info.cdnEndEpoch == 0 && (railId == info.cacheMissRailId || railId == info.cdnRailId)) {
-            info.cdnEndEpoch = endEpoch;
-            emit CDNPaymentTerminated(dataSetId, endEpoch, info.cacheMissRailId, info.cdnRailId);
```
bajtos, line 1626: Are you sure the CDN & cacheMiss payment rails are never terminated? I understand we are using a fixed lockup instead of a payment rate, but I still expect our payment rails to be eventually terminated so that the payer can receive back their remaining funds. When that happens, I think it's good to emit the `CDNPaymentTerminated` event for observability by other components in the system.

Having wrote that, my understanding of the design is limited, I may be wrong.

Let's discuss!

bajtos (changes requested):
Great start! 🤩

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: `railTerminated` is an callback method called by payment rail validator after the payment rail has been terminated. Given that our payment rails are do not have validator this method will never be called after the payment rail termination hence this change.

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: Addressed in https://github.com/FilOzone/filecoin-services/pull/237/commits/5c92d57165c2d812160ee97b7618d2c7646b5ae3

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: CDN rails not configured would mean that either `cdnRailId `or `cacheMissRailId `are set to 0. Payment rail ids won't be removed after (CDN) service has been terminated as we still may want to settle payment rails after termination.

It may be appropriate to emit `FilBeamServiceNotConfigured` error in this example. WDYT?

service_contracts/abi/FilecoinWarmStorageService.abi.json
bajtos: I have a gut feeling that it's more useful to emit the new value, because that's what our offchain indexer worker needs - how much pre-paid egress is remaining?

We could also emit both - the new total amount and the delta.

I don't have a strong opinion about which option is the best.

What I feel strongly about is ensuring the event payload names clearly communicate which case it is (new total vs. delta).

service_contracts/src/FilecoinWarmStorageService.sol
bajtos: Now I see that this is the same case as we are discussing in https://github.com/FilOzone/filecoin-services/pull/237#discussion_r2394795541.

I think we can close this thread as resolved 👍🏻

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,18 +976,91 @@ contract FilecoinWarmStorageService is
...
+     * @param dataSetId The ID of the data set
+     * @param cdnAmount Amount to add to CDN rail lockup
+     * @param cacheMissAmount Amount to add to cache miss rail lockup
+     */
+    function topUpCDNPaymentRails(uint256 dataSetId, uint256 cdnAmount, uint256 cacheMissAmount) external {
+        DataSetInfo storage info = dataSetInfo[dataSetId];
+        require(info.pdpRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        // Check authorization - only payer can top up
+        require(msg.sender == info.payer, Errors.CallerNotPayer(dataSetId, info.payer, msg.sender));
```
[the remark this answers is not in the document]
pyropy: Sadly I think that's not possible due to limitations of Filecoin Pay. Currently lockup can be changed only for a payment rail that's set up between payer and payee (beneficiary). Enabling for anyone to top-up payment rails would enable an attack vector where one could lockup all user funds into a payment rail.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,18 +976,91 @@ contract FilecoinWarmStorageService is
...
+        require(info.cacheMissRailId != 0, Errors.InvalidDataSetId(dataSetId));
+        require(info.cdnRailId != 0, Errors.InvalidDataSetId(dataSetId));
+
+        Payments payments = Payments(paymentsContractAddress);
+
+        if (cdnAmount > 0) {
+            // Get current lockup and increment by the passed amount
+            Payments.RailView memory cdnRail = payments.getRail(info.cdnRailId);
+            // Ensure the rail is not terminated
+            require(cdnRail.endEpoch == 0, Errors.CDNPaymentAlreadyTerminated(dataSetId));
```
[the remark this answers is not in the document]
pyropy: [It's only possible to reduce fixed lockup on a terminated payment rail](https://github.com/FilOzone/filecoin-pay/blob/6a9c4896b4d9f37575899ecc7d9df4414dfaf561/src/Payments.sol#L889), hence the check.

service_contracts/abi/FilecoinWarmStorageService.abi.json
pyropy: @bajtos `topUpCDNPaymentRails` has been updated so that it emits total amount locked up inside payment rails `event CDNPaymentRailsToppedUp(uint256 indexed dataSetId, uint256 totalCdnLockup, uint256 totalCacheMissLockup)`

service_contracts/src/FilecoinWarmStorageService.sol
juliangruber: ping @bajtos 

> How did you come up with the 7:3 split?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -572,29 +580,35 @@ contract FilecoinWarmStorageService is
         uint256 cdnRailId = 0;
 
         if (hasMetadataKey(createData.metadataKeys, METADATA_KEY_WITH_CDN)) {
+            // Create cache-miss payment rail with 0.3 USDFC lockup
```
juliangruber, line 583:
```suggestion
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -572,29 +580,35 @@ contract FilecoinWarmStorageService is
...
-                address(this), // this contract acts as the arbiter
+                address(0), // no validator
                 0, // no service commission
-                address(this)
+                address(this) // controller
             );
             info.cacheMissRailId = cacheMissRailId;
             railToDataSet[cacheMissRailId] = dataSetId;
-            payments.modifyRailLockup(cacheMissRailId, DEFAULT_LOCKUP_PERIOD, 0);
+            // Set lockup with 0.3 USDFC
```
juliangruber, line 594:
```suggestion
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -572,29 +580,35 @@ contract FilecoinWarmStorageService is
...
+            payments.modifyRailLockup(cacheMissRailId, DEFAULT_LOCKUP_PERIOD, DEFAULT_CACHE_MISS_LOCKUP_AMOUNT);
 
+            // Create CDN payment rail with 0.7 USDFC lockup
```
juliangruber, line 597:
```suggestion
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -572,29 +580,35 @@ contract FilecoinWarmStorageService is
...
-                address(this), // this contract acts as the arbiter
+                address(0), // no validator
                 0, // no service commission
-                address(this)
+                address(this) // controller
             );
             info.cdnRailId = cdnRailId;
             railToDataSet[cdnRailId] = dataSetId;
-            payments.modifyRailLockup(cdnRailId, DEFAULT_LOCKUP_PERIOD, 0);
+            // Set lockup with 0.7 USDFC
```
juliangruber, line 608:
```suggestion
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -637,13 +651,7 @@ contract FilecoinWarmStorageService is
         // Check if the data set's payment rails have finalized
         require(
             info.pdpEndEpoch != 0 && block.number > info.pdpEndEpoch,
-            Errors.PaymentRailsNotFinalized(dataSetId, info.pdpEndEpoch, info.cdnEndEpoch)
-        );
-
-        // Check CDN payment rail: either no CDN configured (cdnEndEpoch == 0) or past CDN end epoch
-        require(
-            info.cdnEndEpoch == 0 || block.number > info.cdnEndEpoch,
```
juliangruber, line 645: why don't we need to check `cdnEndEpoch` any more?

service_contracts/src/FilecoinWarmStorageService.sol
juliangruber: ping @pyropy

service_contracts/src/FilecoinWarmStorageService.sol
juliangruber: Should we fetch the payment rail struct and check whether it's ended? As below in topUpCDNPaymentRails

juliangruber (changes requested):

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: `cdnEndEpoch` was set by the validator contract after the payment rail termination. Since our payment rails don't have validator anymore we're not able to set the `cdnEndEpoch` anymore and hence that's why we've removed it. 

Take a look at this [comment](https://github.com/FilOzone/filecoin-services/pull/237#discussion_r2394829501) to get more some more context.

service_contracts/src/FilecoinWarmStorageService.sol
pyropy: > Do you mean to emit the error and continue with the rest of the function, or to revert with that error?

I've meant revert with that error.

Kubuxu (approved):

service_contracts/src/FilecoinWarmStorageService.sol
bajtos: We have seen 80% cache-hit rates, but I don't think that will be true for all users - as the volume of data grows, our bot will no longer keep most of the content warm in cache and the cache-hit ratio will decline. 

Now when there is a cache-miss, the user pays X/TiB for cache-miss request to the SP and X/TiB for the CDN egress. When there is cache-hit, the user pays only X/TiB for the CDN egress.

If we assumed a 50% cache-hit ratio, the user should put 33.3% towards SP payments and 66.7% towards CDN payments.

I rounded those two numbers to 30% and 70%.

Having wrote that, I don't think there is a single perfect split that will work for everyone, so let's start with some value and iterate from there. 
- I am fine to use a different split, E.g. 80% for CDN and 20% for SPs if we expect higher cache-hit rations, or 60% for CDN and 40% for SPs if we expect lower cache-hit rations, or something else.
- **Is it possible to change that split after the smart contract is deployed? I think this is the most important decision to make.**

bajtos (approved):
Other than the discussion around `settleCDNPaymentRails` API, I think the pull request looks great now!

juliangruber (approved):

bajtos (approved):

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -83,6 +83,8 @@ contract FilecoinWarmStorageService is
 
     event ViewContractSet(address indexed viewContract);
 
+    event CDNPaymentRailsToppedUp(uint256 indexed dataSetId, uint256 totalCdnLockup, uint256 totalCacheMissLockup);
```
wjmelements, line 86: Do we really need an event for this?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,10 +982,46 @@ contract FilecoinWarmStorageService is
...
+    /**
+     * @notice Settles CDN payment rails with specified amounts
+     * @dev Only callable by FilCDN (Operator) contract
+     * @param dataSetId The ID of the data set
+     * @param cdnAmount Amount to settle for CDN rail
+     * @param cacheMissAmount Amount to settle for cache miss rail
+     */
+    function settleFilBeamPaymentRails(uint256 dataSetId, uint256 cdnAmount, uint256 cacheMissAmount)
+        external
+        onlyFilBeamController
```
wjmelements, line 994: maybe the filbeamcontroller should be the cdn rail operator so it doesn't have to call through like this?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -978,10 +982,46 @@ contract FilecoinWarmStorageService is
...
+            payments.modifyRailPayment(info.cdnRailId, 0, cdnAmount);
+        }
+
+        if (cacheMissAmount > 0) {
+            payments.modifyRailPayment(info.cacheMissRailId, 0, cacheMissAmount);
+        }
+    }
+
+    /**
+     * @notice Allows users to add funds to their CDN-related payment rails
```
wjmelements, line 1014: I don't think that's what this function does

wjmelements: ok I see it is because you are using one-time payments. Maybe you should be using rates and IValidator instead.

wjmelements: > usage-based payment model

ok

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: The only problem with this would be that users would need to approve additional operators.

Issue #235, opened 2025-09-25
add pdpDatasetId to getClientDataSets


hugomrdias:
Can we have the global dataset id in `getClientDataSets` so we dont have to make another call to get it ?

jennijuju:
can we not use subgraph for it?

jennijuju:
@LexLuthr this is one of the tickets mentioned!

rvagg:
Related to discussion in https://github.com/FilOzone/filecoin-services/issues/234

PR #242, opened 2025-09-26
feat: update getClientDataSets to return dataSetId

LexLuthr:
Fixes https://github.com/FilOzone/filecoin-services/issues/235

This PR adds a new struct DataSetInfoView and returns the same for getClientDataSets. The new struct is a copy of DataSetInfo with additional parameter `uint256 dataSetId`.


rjan90:
Needs a `make update-abi` to be ran for getting the "check ABI changes" to pass

Issue #241, opened 2025-09-26
remove dataset creation fee


jennijuju:
The current implementation of the dataset creation fee makes it challenging to accurately estimate actual storage costs. One of the main reasons for introducing this fee was to prevent storage providers from creating datasets on-chain that are never actually used by clients. However, by enforcing that dataset creation always coincides with the addition of the first data piece (as shown in [this pull request](https://github.com/FilOzone/pdp/pull/201/files)), this concern is addressed. We do not see additional attack factors that would justify the added protocol complexity. Therefore, we are removing the dataset creation fee for the GA release.

jennijuju:
@Chaitu-Tatipamula yah  I think so !

PR #245, opened 2025-09-26
feat: remove dataset creation fee

Chaitu-Tatipamula:
# Remove Dataset Creation Fee

## Summary
Removes the 0.1 USDFC dataset creation fee.

## Changes
- **Removed** `DATA_SET_CREATION_FEE` constant and initialization
- **Updated** `createDataSet()` function to set zero lockup and payment amounts
- **Fixed** tests to expect no fee charges during dataset creation
- **Updated** comments and test names to reflect fee removal

## Rationale
The dataset creation fee was originally introduced to prevent storage providers from creating unused datasets. However, since dataset creation now coincides with adding the first data piece, this concern is addressed without the fee.

## Files Changed
- `src/FilecoinWarmStorageService.sol` - Removed fee logic
- `test/FilecoinWarmStorageService.t.sol` - Updated test expectations

## Testing
- All existing tests pass with updated expectations
- Dataset creation now requires no upfront payment

Issue #241

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -171,7 +171,6 @@ contract FilecoinWarmStorageService is
     address payable private constant BURN_ADDRESS = payable(0xff00000000000000000000000000000000000063);
 
     // Dynamic fee values based on token decimals
-    uint256 private immutable DATA_SET_CREATION_FEE; // 0.1 USDFC with correct decimals
 
     // Token decimals
     uint8 private immutable TOKEN_DECIMALS;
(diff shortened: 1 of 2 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -171,7 +171,6 @@ contract FilecoinWarmStorageService is
     address payable private constant BURN_ADDRESS = payable(0xff00000000000000000000000000000000000063);
 
     // Dynamic fee values based on token decimals
-    uint256 private immutable DATA_SET_CREATION_FEE; // 0.1 USDFC with correct decimals
```
wjmelements, line 174: should also remove the comment

PR #247, opened 2025-09-29
feat: remove provierId from ServiceProviderInfo struct

### LINKED ISSUE #234: Remove providerId from ServiceProviderInfo struct
this does not belong in the struct itself. It is the key! We already have it whenever we lookup this struct!

_Originally posted by @wjmelements in https://github.com/FilOzone/filecoin-services/pull/209#discussion_r2377610965_
            

We want to remove this redundant field from storage. It can still be returned but it doesn't need to be stored.

Chaitu-Tatipamula:
# Remove providerId from ServiceProviderInfo struct

## Summary
Remove `providerId` from `ServiceProviderInfo` struct and introduce `ServiceProviderInfoView` as a view struct to return provider ID with provider info.

## Related Issue
Fixes #234

## Changes
- **Removed** `providerId` field from `ServiceProviderInfo` struct
- **Added** `ServiceProviderInfoView` struct with `providerId` field for external consumption
- **Updated** all getter functions to return `ServiceProviderInfoView`
- **Updated** tests and ABI definitions

## Files Modified
- `ServiceProviderRegistry.sol`
- `ServiceProviderRegistryStorage.sol` 
- `FilecoinWarmStorageService.sol`
- Test files and ABI

### DIFF
--- a/service_contracts/src/ServiceProviderRegistryStorage.sol
+++ b/service_contracts/src/ServiceProviderRegistryStorage.sol
@@ -25,16 +25,6 @@ contract ServiceProviderRegistryStorage {
         bool isActive;
     }
 
-    /// @notice Provider information for API returns
-    struct ServiceProviderInfoView {
-        uint256 providerId; // Provider ID
-        address serviceProvider; // Address that controls the provider registration
-        address payee; // Address that receives payments (cannot be changed after registration)
-        string name; // Optional provider name (max 128 chars)
-        string description; //Service description, ToC, contract info, website..
-        bool isActive;
-    }
-
     /// @notice Product offering of the Service Provider
     struct ServiceProduct {
         ProductType productType;
@@ -59,7 +49,7 @@ contract ServiceProviderRegistryStorage {
     /// @notice Combined provider and product information for detailed queries
     struct ProviderWithProduct {
         uint256 providerId;
-        ServiceProviderInfoView providerInfo;
+        ServiceProviderInfo providerInfo;
         ServiceProduct product;
     }
 
--- a/service_contracts/src/ServiceProviderRegistryStorage.sol
+++ b/service_contracts/src/ServiceProviderRegistryStorage.sol
@@ -23,7 +23,16 @@ contract ServiceProviderRegistryStorage {
         string name; // Optional provider name (max 128 chars)
         string description; //Service description, ToC, contract info, website..
         bool isActive;
-        uint256 providerId; // Unique identifier for the provider
+    }
+
+    /// @notice Provider information for API returns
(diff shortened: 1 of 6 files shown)


service_contracts/src/ServiceProviderRegistryStorage.sol
```diff
@@ -23,7 +23,16 @@ contract ServiceProviderRegistryStorage {
         string name; // Optional provider name (max 128 chars)
         string description; //Service description, ToC, contract info, website..
         bool isActive;
-        uint256 providerId; // Unique identifier for the provider
+    }
+
+    /// @notice Provider information for API returns
+    struct ServiceProviderInfoView {
```
wjmelements, line 29: I don't think this struct belongs in Storage.sol because it isn't related to storage. Put it into the ServiceProviderRegistry.

service_contracts/src/ServiceProviderRegistryStorage.sol
```diff
@@ -50,7 +59,7 @@ contract ServiceProviderRegistryStorage {
     /// @notice Combined provider and product information for detailed queries
     struct ProviderWithProduct {
         uint256 providerId;
-        ServiceProviderInfo providerInfo;
+        ServiceProviderInfoView providerInfo;
```
wjmelements, line 62: This doesn't need InfoView struct because it already has `providerId`.

service_contracts/src/ServiceProviderRegistryStorage.sol
```diff
@@ -23,7 +23,16 @@ contract ServiceProviderRegistryStorage {
...
+    }
+
+    /// @notice Provider information for API returns
+    struct ServiceProviderInfoView {
+        uint256 providerId; // Provider ID
+        address serviceProvider; // Address that controls the provider registration
+        address payee; // Address that receives payments (cannot be changed after registration)
+        string name; // Optional provider name (max 128 chars)
+        string description; //Service description, ToC, contract info, website..
+        bool isActive;
```
wjmelements, line 35: Let's reuse ServiceProviderInfo here, like
```
struct ServiceProviderInfoView {
    uint256 providerId;
    ServiceProviderInfo info;
}
```

wjmelements (changes requested):
My original thought was that this didn't need to be a breaking change, that we could keep the ABI the same. My opinion has changed for three reasons:
1. ProviderWithProduct had redundant providerId.
2. providerId fits more naturally at the beginning of the view struct.
3. Can nest the type ServiceProviderInfo inside ServiceProviderInfoView

I don't think it will be a hard set of changes in synapse-sdk and curio.

PR #249, opened 2025-09-29
perf(ServiceProviderRegistry): getProviderPayee helper for dataSetCreated

### LINKED ISSUE #248: perf(ServiceProviderRegistry): getProviderPayee helper for dataSetCreated
Separate task: We should make a helper to fetch only the payee. I think that's the only field we use from providerInfo here. It's not good to read all of this information from storage, including two strings, and then discard it.

_Originally posted by @wjmelements in https://github.com/FilOzone/filecoin-services/pull/247#discussion_r2388980803_
            

We want to make a helper to fetch the payee address for a service provider to reduce the access footprint of CreateDataSet. Currently we fetch the entire provider info when we only need the payee.

Chaitu-Tatipamula:
## Description

Implements a helper function to fetch only the payee address for service providers, reducing the access footprint of `CreateDataSet`.

## Problem

Currently, `FilecoinWarmStorageService.dataSetCreated()` calls `getProvider(providerId)` which reads the entire `ServiceProviderInfo` struct from storage, including expensive string fields (`name` and `description`), when only the `payee` address is needed.

## Solution

- **Added `getProviderPayee(uint256 providerId)`** in `ServiceProviderRegistry.sol`
  - Returns only the payee address directly from storage
  - Uses `providerExists` modifier for safety

- **Updated `FilecoinWarmStorageService.sol`**
  - Replaced `getProvider()` call with `getProviderPayee()` in `dataSetCreated()`
  - Eliminates intermediate struct creation and string reads

- **Added unit tests**
  - `testGetProviderPayeeReturnsCorrectAddress()` - verifies correct payee returned
  - `testGetProviderPayeeRevertsForInvalidProviderId()` - verifies revert behavior

Fixes #248

### DIFF
--- a/service_contracts/src/ServiceProviderRegistry.sol
+++ b/service_contracts/src/ServiceProviderRegistry.sol
@@ -484,6 +484,13 @@ contract ServiceProviderRegistry is
         return ServiceProviderInfoView({providerId: providerId, info: provider});
     }
 
+    /// @notice Get only the payee address for a provider
+    /// @param providerId The ID of the provider
+    /// @return payee The payee address
+    function getProviderPayee(uint256 providerId) external view providerExists(providerId) returns (address payee) {
+        return providers[providerId].payee;
+    }
+
     /// @notice Get product data for a specific product type
     /// @param providerId The ID of the provider
     /// @param productType The type of product to retrieve
--- a/service_contracts/abi/ServiceProviderRegistry.abi.json
+++ b/service_contracts/abi/ServiceProviderRegistry.abi.json
@@ -826,6 +826,25 @@
   },
   {
     "type": "function",
+    "name": "getProviderPayee",
+    "inputs": [
+      {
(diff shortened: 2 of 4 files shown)


service_contracts/src/ServiceProviderRegistry.sol
```diff
@@ -484,6 +484,13 @@ contract ServiceProviderRegistry is
         return ServiceProviderInfoView({providerId: providerId, info: provider});
     }
 
+    /// @notice Get only the payee address for a provider
+    /// @param providerId The ID of the provider
+    /// @return payee The payee address
+    function getProviderPayee(uint256 providerId) external view providerExists(providerId) returns (address payee) {
```
wjmelements, line 490: I think `providerExists` might also be suboptimal, but there are probably more urgent matters to fix

wjmelements (approved):

PR #263, opened 2025-10-06
feat: Automate fetching of USDFC contract ABI

pali101:
This change is a precursor to solve FilOzone/synapse-sdk#244. The SDK requires the full USDFC ABI to integrate features that use extensions like ERC-2612 and ERC-3009. This PR provides a reliable, automated way to source that ABI.

## Problem

Previously, we used a generic `ERC20` ABI which lacks support for the `ERC-2612` and `ERC-3009` specific extensions. This required manual ABI additions and could lead to discrepancies, blocking integration in `synapse-sdk`.

## Solution

- A new script, `tools/fetch-usdfc-abi.sh`, is added. It resolves the USDFC proxy address to its underlying implementation contract and fetches the correct ABI
- The `Makefile` is updated to integrate the script. The USDFC ABI can be fetched on-demand with make `abi/Usdfc.abi.json` and is also automatically updated with all other contracts when running `make update-abi`
- The most recent USDFC ABI has been checked in to reflect this change

This streamlines the process of keeping the ABI in sync with the deployed contract and ensures all necessary function definitions are available.

### DIFF
--- a/service_contracts/Makefile
+++ b/service_contracts/Makefile
@@ -5,6 +5,7 @@ RPC_URL ?=
 KEYSTORE ?=
 PASSWORD ?=
 CHALLENGE_FINALITY ?=
+USDFC_PROXY_ADDRESS ?= 0x80B98d3aa09ffff255c3ba4A241111Ff1262F045
 
 # Default target
 .PHONY: default
@@ -134,7 +135,7 @@ ABI_CONTRACTS := \
 	SessionKeyRegistry
 
 # Generate ABI file targets
-ABI_FILES := $(addprefix abi/,$(addsuffix .abi.json,$(ABI_CONTRACTS)))
+ABI_FILES := $(addprefix abi/,$(addsuffix .abi.json,$(ABI_CONTRACTS))) abi/Usdfc.abi.json
 
 # Define a template for ABI extraction; we use a template approach instead of
 # defining a global pattern rule because patterns for files that don't exist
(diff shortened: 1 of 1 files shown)


service_contracts/tools/fetch-usdfc-abi.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+
+#
+# Fetches the ABI for a Filecoin proxy contract's implementation and saves it.
+#
```
rvagg, line 5:
```suggestion
# Fetches the ABI for a Filecoin mainnet proxy contract's implementation and saves it.
#
```

service_contracts/tools/fetch-usdfc-abi.sh
```diff
@@ -0,0 +1,85 @@
...
+  die "Failed to fetch data for implementation contract."
+
+# 3. Extract, parse, and save the ABI.
+# This single command extracts the .abi string, parses it from a string into
+# clean JSON, and saves it. It will fail if any step is unsuccessful.
+echo "$IMPL_RESPONSE" | jq '.abi | fromjson' > "$OUTPUT_PATH" || {
+  # Clean up the potentially empty/corrupt file on failure.
+  rm -f "$OUTPUT_PATH" 2>/dev/null
+  die "Failed to parse ABI. The '.abi' field may be missing, null, or not valid JSON."
+}
```
rvagg, line 83: This should be more graceful - you could send it to a temporary file and then move it to OUTPUT_PATH, but the way this is now, a failure to fetch will leave you with a deleted ABI file in your repo.

BigLep:
@hugomrdias : what are the next steps here?

hugomrdias:
We discussed this and we will keep the static abi for ERC20 that we have already to avoid adding more moving parts.

Issue #226, opened 2025-09-16
add PeerID for PDP offering metadata


jennijuju:
We considered to derive peerID for IPNI using SP's owner wallet address (aka the operational address) - which comes with a downside - whenever the SP wants to rotate their operational keys (which will lead to a new peerID), all their ads need to be updated with new peerID and that is an expensive operation (basically establish a new ad chain), especially for large providers. PeerID really doesn't have any binding relationship with a dataset or a piece perse - client will still retrieve from the the sp's service url

So to keep the peerID relatively stable for reducing IPNI ops, we could use a `pdpOffering.IPNIPeerId` .

rvagg:
https://github.com/filecoin-project/builtin-actors/blob/88c3befb5c0a49d4e61af38e934c1ae64629885b/runtime/src/runtime/policy.rs#L305-L307

Miner actor has a max byte length of 128 for this.

jennijuju:
@rvagg can you please document and share what changes your mind, please.

rvagg:
@jennijuju:

> We should be able to use this to confirm IPNI advertisements were done.

It's the final step in the Filecoin Pin flow, and we have nothing to tell us that it's done, and we know there's going to be a `<some number of seconds or minutes>` delay, both because of the Curio task system delay _and_ the lazy advertisement pull from IPNI (we don't get to push advertisements, we plead with IPNI to come and get them from us).

So in any Filecoin Pin use, we have this moment of unavailability that's basically impossible to even poll to figure out. There are two options that I see, both of which we could pursue:

1. Poll cid.contact directly using our provider's peer ID and root CID to see if it's there, when it is, assume they all are (that's good enough for now)
2. Create a new API on Curio to ask it if it's notified IPNI .. which still isn't quite good enough to tell us about whether IPNI is listing our CIDs.

So option 1 seems to be the best we can do, and we need to get the peer ID from somewhere.

https://github.com/FilOzone/filecoin-services/pull/266

PR #266, opened 2025-10-07
feat!(ServiceProviderRegistry): add ipniPeerId to PDPOffering

rvagg:
Closes: https://github.com/FilOzone/filecoin-services/issues/226

@LexLuthr we're only going to need one of these, right? We're not going to use a second one if we advertise PieceCID as well?

I'm thinking that we can use this to do a cross-reference with cid.contact that the SP has actually published _something_. A check for the root CID will have to be enough, and is most likely sufficient for most use cases anyway.

### DIFF
--- a/service_contracts/abi/ServiceProviderRegistry.abi.json
+++ b/service_contracts/abi/ServiceProviderRegistry.abi.json
@@ -426,6 +426,11 @@
             "name": "paymentTokenAddress",
             "type": "address",
             "internalType": "contract IERC20"
+          },
+          {
+            "name": "ipniPeerId",
+            "type": "bytes",
+            "internalType": "bytes"
           }
         ]
       },
@@ -1252,6 +1257,11 @@
             "name": "paymentTokenAddress",
             "type": "address",
             "internalType": "contract IERC20"
+          },
+          {
+            "name": "ipniPeerId",
+            "type": "bytes",
+            "internalType": "bytes"
           }
         ]
       },
--- a/service_contracts/test/ServiceProviderRegistryFull.t.sol
+++ b/service_contracts/test/ServiceProviderRegistryFull.t.sol
@@ -1939,7 +1939,9 @@ contract ServiceProviderRegistryFullTest is Test {
         bytes memory encodedUpdatedData = abi.encode(updatedData);
 
         vm.prank(provider1);
-        registry.updateProduct(ServiceProviderRegistryStorage.ProductType.PDP, encodedUpdatedData, emptyKeys, emptyValues);
+        registry.updateProduct(
+            ServiceProviderRegistryStorage.ProductType.PDP, encodedUpdatedData, emptyKeys, emptyValues
+        );
 
         // Verify update
         (ServiceProviderRegistryStorage.PDPOffering memory updatedStored,,) = registry.getPDPService(providerId);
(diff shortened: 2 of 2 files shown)


LexLuthr:
> Closes: #226
> 
> @LexLuthr we're only going to need one of these, right? We're not going to use a second one if we advertise PieceCID as well?
> 
> I'm thinking that we can use this to do a cross-reference with cid.contact that the SP has actually published _something_. A check for the root CID will have to be enough, and is most likely sufficient for most use cases anyway.

Yes, we only need 1. MKv2 uses a single peerID per provider to publish all ads. PDPV0 only seem to publish IPFS so this will work for both.

service_contracts/src/ServiceProviderRegistryStorage.sol
```diff
@@ -44,6 +44,7 @@ contract ServiceProviderRegistryStorage {
         uint256 minProvingPeriodInEpochs; // Minimum proving period in epochs
         string location; // Geographic location of the service provider
         IERC20 paymentTokenAddress; // Token contract for payment (IERC20(address(0)) for FIL)
+        bytes ipniPeerId; // IPNI peer ID (max 128 bytes, can be empty)
```
LexLuthr, line 47: Should be just use string here instead of byte? Less likely to use bytes anywhere outside of libp2p and IPNI internal.  String also does not require extra manipulation to do RPC calls to IPNI and is human readable.

LexLuthr (approved):

service_contracts/src/ServiceProviderRegistryStorage.sol
rvagg: Actually maybe not. I need to sleep on this but the byte form is the only stable form, string form could be any base encoded version of it. There's also the complexity with maximum size. 127 is an easy maximum, it matches the miner actor. But choosing a max for strings is awkward, probably don't need to support base256emoji, base32 gets us somewhere over 200 chars, base52btc as a typical CIDv0 is smaller but can we mandate that? Two different string forms can resolve to the same byte form, so matching on strings isn't ideal.
But .. subgraphs and all that.

LexLuthr: Definitely a few tradeoffs. As for string, we can ensure a single type of conversion. Curio will always use peerID.String() function. SDK can probably do something similar. Remove the string complexity but keep the human readable and API query aspects.

wjmelements:
> Although technically I don't think this is breaking and could be upgraded in the implementation and we wouldn't need a new contract deployment, I think that's right @wjmelements? If we add to the end then it's an upgrade and existing clients can still abi.decode into the rest of the struct and just ignore this?

It breaks the ABI of `encodePDPOffering` because it changes the 4byte signature. I'm not sure if it breaks `decodePDPOffering` et al decoding; it would depend on if the library ignores extra data. It should be fine on-chain though.

Appending to structs is generally upgrade-safe for proxy storage layouts. However, essentially all fields are permanent after that.

wjmelements:
> Appending to structs is generally upgrade-safe for proxy storage layouts. However, essentially all fields are permanent after that.

Upon further review, this isn't appending to a struct but to encoded `bytes storage`. So this would break the product data.

service_contracts/src/ServiceProviderRegistryStorage.sol
rvagg: agreed to go with bytes

PR #287, opened 2025-10-13
Reduce decodePDPOffering

wjmelements:
Reviewer @redpanda-f
We have this helper function, but we haven't been using it.

Reduces ServiceProviderRegistry codesize 21,615 -> 21,367
#### Changes
* Emit productData in events instead of serviceUrl.
* Reduce decodePDPOffering by not decoding product data outside of validation


silent-cipher:
Yes, I can do local abi decoding in subgraph- will remove use of 'decodePDPOffering' in subgraph

PR #294, opened 2025-10-16
feat!(ServiceProviderRegistry): emit raw productData

wjmelements:
Reviewer @rvagg
Supercedes https://github.com/FilOzone/filecoin-services/pull/287
As suggested [here](https://github.com/FilOzone/filecoin-services/pull/287#discussion_r2427894550).
We want this change because it will make the event more portable to/forward-compatible with future product types.
This change slightly reduces ServiceProviderRegistry codesize (21,615 -> 21,239).
This is a breaking change because the `serviceUrl` arg is removed from the Product events.
However, the `serviceUrl` can still be obtained from the product data if necessary.
#### Changes
* remove encodePDPOffering and decodePDPOffering
* use msg.sender for serviceProvider in Product events
* remove PDP specific field serviceUrl from Product events; emit all raw product data


rvagg (approved):
eggzelent

(take note @silent-cipher this will impact subgraph for the next upgrade)

wjmelements:
> May I push those changes here or raise another pr?

Please raise another PR

PR #299, opened 2025-10-17
chore: migrate from git submodules to npm workspaces monorepo structure

### LINKED ISSUE #229: Proposal: monorepo & no git submodules

Chaitu-Tatipamula:
# Migrate from Git Submodules to NPM Workspaces Monorepo

## Summary
Migrates the Filecoin Services project from git submodules to a unified npm workspaces monorepo structure to improve developer experience and simplify dependency management.

## Changes
- **Remove git submodules** and `service_contracts/` directory
- **Consolidate packages** into unified monorepo with npm workspaces
- **Centralize dependencies** using npm for all packages except `fvm-solidity`
- **Single Foundry project** with unified build/test system
- **Consolidated Makefile** with 50+ targets for all development needs
- **Updated GitHub workflows** to work with new monorepo structure
- **ABI extraction system** compatible with new structure

Resolves #229

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ broadcast/
 node_modules/
 
 # Foundry dependencies
-lib/
+/lib
 
 # Ignore IDEs
 .idea
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/fvm-solidity"]
+	path = lib/fvm-solidity
+	url = https://github.com/filecoin-project/fvm-solidity
--- a/packages/warm-storage/src/lib/FilecoinWarmStorageServiceLayout.sol
+++ b/packages/warm-storage/src/lib/FilecoinWarmStorageServiceLayout.sol
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+pragma solidity ^0.8.20;
+
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any changes will be lost.
+// Generated with tools/generate_storage_layout.sh
+
+bytes32 constant MAX_PROVING_PERIOD_SLOT = bytes32(uint256(0));
+bytes32 constant CHALLENGE_WINDOW_SIZE_SLOT = bytes32(uint256(1));
+bytes32 constant SERVICE_COMMISSION_BPS_SLOT = bytes32(uint256(2));
+bytes32 constant PROVEN_PERIODS_SLOT = bytes32(uint256(3));
+bytes32 constant PROVING_ACTIVATION_EPOCH_SLOT = bytes32(uint256(4));
+bytes32 constant PROVING_DEADLINES_SLOT = bytes32(uint256(5));
+bytes32 constant PROVEN_THIS_PERIOD_SLOT = bytes32(uint256(6));
+bytes32 constant DATA_SET_INFO_SLOT = bytes32(uint256(7));
+bytes32 constant CLIENT_DATA_SET_IDS_SLOT = bytes32(uint256(8));
+bytes32 constant CLIENT_DATA_SETS_SLOT = bytes32(uint256(9));
+bytes32 constant RAIL_TO_DATA_SET_SLOT = bytes32(uint256(10));
+bytes32 constant DATA_SET_METADATA_SLOT = bytes32(uint256(11));
+bytes32 constant DATA_SET_METADATA_KEYS_SLOT = bytes32(uint256(12));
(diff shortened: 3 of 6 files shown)


wjmelements:
We have a nice modular codebase. Submodules are great. Monorepos are a burden in minor ways.

1. The tests are all run for everything every time. This wastes money and developer time. Developers may workaround by running fewer tests, but CI will not.
2. Reusing our modular contracts such as FilecoinPayV1 for other services is more difficult with monorepos. Their build and install times are longer, though they can sometimes workaround by [undoing this change on their own fork](https://stackoverflow.com/questions/5303496/how-to-change-a-git-submodule-to-point-to-a-subfolder).
3. Modular code in monorepos degrades into interdependent code, causing import cycles.
4. Monorepos have unbounded scope and so absorb everything they touch. There is no limiting principle preventing us, for example, from bringing synapse-sdk in here too. Our future services would probably also accumulate here.
5. Migration eliminates the `git blame` and `git log` which are helpful for digging up context and rationale.
6. The likelihood of merge conflicts and other incompatible changes is higher when the whole team is working at the same time on the same repo. At scale this can create a chokepoint and a merge backlog that can grow faster than it takes to run the tests.

I'm not convinced that the benefits outweigh the costs.

Issue #300, opened 2025-10-17
Anyone can call validatePayment, which emits the PaymentArbitrated event


wjmelements:
`FilecoinWarmStorageService.validatePayment` does not authenticate `msg.sender`, so anyone can call it. This would be fine if the method was `view`, which it almost is. However, it emits `PaymentArbitrated`. Consequently, this log is noise because anyone can cause it to be emitted.

Two possible solutions:
1. Remove `PaymentArbitrated` event and restrict `FilecoinWarmStorageService.validatePayment` to `view`.
2. Require `msg.sender` in `FilecoinWarmStorageService.validatePayment` is `paymentsContractAddress`.

I'm not especially attached to this `PaymentArbitrated` event. It would make more sense for it to be emitted by Payments if anywhere, but maybe it doesn't need to be emitted at all. I would prefer to remove it because I don't think anything is using it.

PR #301, opened 2025-10-17
fix: remove PaymentArbitrated

wjmelements:
Reviewers @rvagg @zenground0
Fixes https://github.com/FilOzone/filecoin-services/issues/300
If we remove this event, this IValidator method can be `view`.
I don't believe the event is used anywhere.
#### Changes
* rm PaymentArbitrated event

### DIFF
--- a/subgraph/API.md
+++ b/subgraph/API.md
@@ -335,7 +335,6 @@ query AllRails($first: Int, $skip: Int) {
     operator
     arbiter
     paymentRate
-    settledUpto
     endEpoch
     queueLength
     dataSet {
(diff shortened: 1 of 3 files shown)


silent-cipher:
I agree with ripping it out. If we do that, we should also remove the following fields from `rail` as well:
```
  settledUpto: BigInt!
  settledAmount: BigInt!
  totalFaultedEpochs: BigInt!
```
Instead of keeping these as falsy or unused values, it’s cleaner to remove them entirely.

If someone needs anything related to rails, they can always fetch that from the [payments subgraph](https://github.com/FilOzone/filecoin-pay-explorer/blob/main/packages/subgraph/schemas/schema.v1.graphql).

Issue #291, opened 2025-10-16
Any SPs can join FWSS; ApprovedProviders are FilOz;s approved providers for Synapse


BigLep:
## Problems to solve
1. Make it self-service for an SP to test their participation in FWSS
2. Have a clear self-service way for FilOz to denote the SPs that are demonstrating sufficient "performance/reliabilty" to be used in Synpase for projects like Filecoin Pin

## Proposed solution
Part 1: make it so any SP can self-service join FWSS
Part 2: "Approved Providers" now becomes the "whitelisted/blessed" set of SPs that should be considered by Synapse during Data Set creation.  

## Merits of this approach
This addresses the "problems to solve" above.  An SP can self-service join FWSS and then can test their setup by providing an override providerId to a tool like `filecoiin-pin` to test things locally, all without FilOz involvement.

In addition, there are no code changes in Synapse needed since Synapse today already filters down to "Approved Providers".  Whereas today the "Approved Providers" list has SPs manifesting reliability problems, with this change, it will be more tightly curated to a handful of SPs that  have a reliable track records so we have a good foundation to build on for the other new software being built on top like filecoin-pin.

## Implementation Notes
1. The change to FilecoinWarmStorageService.sol is straightforward.  The diff is below.
2. Tests will need to be updated.
3. [dealbot](https://github.com/FilOzone/dealbot) will need to be updated to test all FWSS SPs, not just the approved ones so that "scores" can be generated for SPs before they become approved, as a sufficiently high "score" may even become a requirement for becoming "approved".

### Contact change
```diff
diff --git a/service_contracts/src/FilecoinWarmStorageService.sol b/service_contracts/src/FilecoinWarmStorageService.sol
index 2d8bb47..2c1603f 100644
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -521,9 +521,6 @@ contract FilecoinWarmStorageService is
 
         require(providerId != 0, Errors.ProviderNotRegistered(serviceProvider));
 
-        // Check if provider is approved
-        require(approvedProviders[providerId], Errors.ProviderNotApproved(serviceProvider, providerId));
-
         address payee = serviceProviderRegistry.getProviderPayee(providerId);
 
         require(
@@ -962,9 +959,6 @@ contract FilecoinWarmStorageService is
         // Check if provider is registered
[...]

PR #302, opened 2025-10-19
feat: Make FWSS permissionless by removing provider approval checks

pali101:
This PR is a work-in-progress to address a part of FilOzone/dealbot#30.

This PR removes the on-chain requirement for a Service Provider (SP) to be in the `approvedProviders` list to create or receive datasets.

## Changes
- `FilecoinWarmStorageService.sol`:
    - Removed the `require(approvedProviders[...])` check from `dataSetCreated`.
    - Removed the `require(approvedProviders[...])` check from `storageProviderChanged`.
- `ProviderValidation.t.sol`:
   - Modified `testProviderRegisteredButNotApproved`: The test no longer expects a revert. It now asserts that dataset creation succeeds for any registered provider.
- `FilecoinWarmStorageServiceOwner.t.sol`:
   - Renamed `testStorageProviderChangedRevertsForUnapprovedProvider` to `testStorageProviderChangedSucceedsForAnyRegisteredProvider` and updated the test logic to assert that a `storageProviderChanged` operation succeeds when transferring to a registered (but not explicitly approved) provider.

## Breaking change

The contract no longer enforces that SPs must be "approved" to create datasets or receive dataset transfers. Any SP in the `ServiceProviderRegistry` is now eligible. Clients that rely on a curated list of SPs must now implement this filtering logic themselves by querying the `getApprovedProviders` view function.

## To-do

- [x] Update dealbot to test all FWSS SPs, not just the approved list (as noted in FilOzone/dealbot#30): Done via FilOzone/dealbot#29

### DIFF
--- a/service_contracts/test/ProviderValidation.t.sol
+++ b/service_contracts/test/ProviderValidation.t.sol
@@ -158,7 +158,7 @@ contract ProviderValidationTest is MockFVMTest {
         vm.mockCall(address(0x01), bytes(hex""), abi.encode(client));
 
         vm.prank(provider1);
-        // Dataset creation shouldn't require provider be approved 
+        // Dataset creation shouldn't require provider be approved
         uint256 dataSetId = pdpVerifier.createDataSet(PDPListener(address(warmStorage)), extraData);
 
         // Verify the dataset was created


service_contracts/test/ProviderValidation.t.sol
```diff
@@ -142,8 +156,11 @@ contract ProviderValidationTest is MockFVMTest {
         vm.mockCall(address(0x01), bytes(hex""), abi.encode(client));
 
         vm.prank(provider1);
-        vm.expectRevert(abi.encodeWithSelector(Errors.ProviderNotApproved.selector, provider1, 1));
-        pdpVerifier.createDataSet(PDPListener(address(warmStorage)), extraData);
+        // Dataset creation should now succeed for any registered provider
```
rvagg:
```suggestion
        // Dataset creation should succeed for any registered provider
```

also please add a note above the test `function` or just inside it that says that this operation is expected to pass, approval is not required to perform onboarding actions

rvagg:
I'd like @Kubuxu || @ZenGround0 to cast an eye over this and original issue @ https://github.com/FilOzone/filecoin-services/issues/291 to make sure I'm not missing any good reason not to unlock this.

The idea is that the SDK will pull from and select from the approved provider list, but being in that list won't stop you from being usable. You could still explicitly select any provider in the provider list to use (with ID or address provided to the SDK). Providers could even test their own set up as part of PDP onboarding, and we can test all providers in the registry with dealbot. You could also make "direct" deals with SPs if you want to bypass the list with an override.

Kubuxu (approved):

wjmelements (commented):
Can we also remove `approvedProviders`?

wjmelements:
> Can we also remove `approvedProviders`?

> Clients that rely on a curated list of SPs must now implement this filtering logic themselves by querying the getApprovedProviders view function.

I suppose this is fine. We were also thinking of moving these concepts into the SP Registry.

service_contracts/test/ProviderValidation.t.sol
```diff
@@ -142,8 +158,11 @@ contract ProviderValidationTest is MockFVMTest {
         vm.mockCall(address(0x01), bytes(hex""), abi.encode(client));
 
         vm.prank(provider1);
-        vm.expectRevert(abi.encodeWithSelector(Errors.ProviderNotApproved.selector, provider1, 1));
-        pdpVerifier.createDataSet(PDPListener(address(warmStorage)), extraData);
+        // Dataset creation shouldn't require provider be approved 
+        uint256 dataSetId = pdpVerifier.createDataSet(PDPListener(address(warmStorage)), extraData);
```
rvagg:
```suggestion
        // Dataset creation shouldn't require provider be approved
        uint256 dataSetId = pdpVerifier.createDataSet(PDPListener(address(warmStorage)), extraData);
```

Issue #297, opened 2025-10-17
Review PDPOffering fields


wjmelements:
Many of these fields in the current schema can be capability keys. Capability keys are flexible but the typing of productData (eg PDPOffering) cannot safely change after launch. We should be sure about which fields are universally mandatory and make the others capabilities.

We should also increase the capabilities limit somewhat.

wjmelements:
Via https://github.com/FilOzone/filecoin-services/pull/266, we are also considering ipni peer ID.

wjmelements:
> We discussed it from product side and prefer to keep the pricing as a mandatory field for pdp offering - if the SP doesn’t want to set it, they can pro-claim it to be 0. This gives the network more price discoverability without needing to push for standards, which may be harder for tooling adoption.

"PerMonth" is a pretty bad standard because months have 28-31 days.. Needs a calendar library to support it. Price per epoch, the very basis of Payments, would change month to month. Unless we want to define a month to be 1/12 of a year of course.

jennijuju:
> > We discussed it from product side and prefer to keep the pricing as a mandatory field for pdp offering - if the SP doesn’t want to set it, they can pro-claim it to be 0. This gives the network more price discoverability without needing to push for standards, which may be harder for tooling adoption.
> 
> "PerMonth" is a pretty bad standard because months have 28-31 days.. Needs a calendar library to support it. Price per epoch, the very basis of Payments, would change month to month. Unless we want to define a month to be 1/12 of a year of course.

Let’s put it per epoch or per day to address your concern, which I see the point, and services/app can derive any monthly / annual price from it if needed

wjmelements:
> Currently capabilities are awkward to fetch, you can get keys but have to follow up to get values. If we move more things to capabilities then I'd like nicer ways to eth_call to get everything in one go

On-chain though, it is much cheaper to fetch from a capability than the PDPOffering, because if want to read one item from the PDPOffering, you have to load the whole thing.

jennijuju:
@wjmelements other services my have  different definition. I think changing to per day is a good middle point to achieve both onchain sp offering pricing discoverability & allow service creators to leverage it in their own billing/pricing if needed

wjmelements:
An `extsload` library could also make the `productData` fields individually accessible.

PR #308, opened 2025-10-20
perf!(ServiceProviderRegistry): Bloom Schema

### LINKED ISSUE #307: perf(ServiceProviderRegistry): efficient productData
> @wjmelements we've just had a sync discussion about this and we're both open to changing this even to the point of nuking `PDPOffering` entirely and just having a k/v list with an external schema (i.e. that the SDK and Curio agree on, everything not within that schema is essentially ignored). In the SDK we can filter out SPs that don't have the keys that we need. Listen to the recording of the daily sync for more but my thoughts are something like this: we mainly care about this where we might need information cross-contract and it shouldn't be _too_ expensive to find it. e.g. a hypothetical service contract for a PDP service that only deals in SPs that have `ipniIndexing` turned on, the contract would have to go and figure that out from the registry. But to your point - currently having to decode the blob is a cost as well, so nothing is free here. Possibly also we just convert this one to a `string=bytes` too for efficiency. Right now the main one I care about is `serviceUrl`, but if I have a good way to get the whole set of keys/values in a single call then we can deal with it in the SDK just fine.
> So, permission to radically pull this apart if you want. Maybe an "offering" is simply just a key/value list, with nothing fancy. 

 _Originally posted by @rvagg in [#297](https://github.com/FilOzone/filecoin-services/issues/297#issuecomment-3421692163)_

The current design has two kinds of product attributes:
1. productData, which contained the fields mandated by the productType
2. capabilityKeys, which allowed products to surface additional information extending the standard schema

However, looking up fields from the `bytes storage productData` is inefficient because on-chain consumers must decode the entire bytestring to read one field from it. There are two ways to fix this.

1. An `extsload` pattern could provide a better interface for reading the fields individually without needing to decode them all. We would have one view library per product type.
2. Make every attribute into a capability key. Then we could use bloom filters to enforce the schema for specific product types.

The `extsload` approach works decoding typed data because it is ABI-encoded. We can also make some assumptions about the shape and size of the fields that we cannot make if everything is bytes or string. However this requires a view contract per product type. Also, the optional fields would still be typed as `bytes` or `string`.

The remove productData approach has some nice simplicity to it. Instead of managing two kinds of attributes, there would only be one type. If fields are all `bytes`, some space is wasted for fields such as `uint256` that are not `bytes`, `string`, or `[]`. This freedom could be chaotic onchain if there was some important field that was supposed to be `uint256` but was not.  Products not encoded properly could be filtered-out off-chain, but valid products can be updated to become invalid. This is a problem with all of the capability keys but I raise it here because this approach would make everything into capability keys.

I am currently preferring to remove productData because it simplifies the codebase and does not create more contracts. The benefit of less storage for scalar fields for the `extsload` approach is limited to the ones in `productData`, and I believe there must be at least two of them to break-even.

wjmelements:
Reviewer @rvagg
Closes #307
This significantly reduces the size of ServiceProviderRegistry: 21,290 -> 17,751 (-3539)

#### Motivation
We want to allow all product attributes to be queryable on-chain. Therefore we are removing the encoded productData. Mandatory schema will be loosely enforced on-chain with a bloom filter.

#### Upgrade guide for synapse and curio
1. All `productInfo` previously abi-encoded is removed and everything is now capability key-value store. Keys are `string`; values are `bytes`. This affects `registerProvider`, `addProduct`, and `updateProduct`. Unsigned integers should be encoded big-endian. Addresses should be encoded as `bytes[20]`. Strings should be encoded utf-8. See the examples in `test/PDPOffering.sol`. We don't validate these values on-chain so if they don't look right, throw.
2. `getPDPOffering` is not how you get PDP product info anymore. Now use `getAllProductCapabilities` which will return all keys and values for a product.
3. The `ProviderWithProduct` return type now contains `productCapabilityValues` so you don't have to fetch them separately.
4. `getProductCapabilities` no longer returns the `exists` `bool` array. This array was misleading because it only indicated whether the bytes were empty.
5. `getProductCapability` is removed. Use `productCapabilities` instead, which only differs in not having `exists`.
6. `updatePDPServiceWithCapabilities` is removed. Use `updateProduct`.
7. Capability values cannot be empty. Exclude the key to signal that the product does not have the capability.
8. `getProvidersByProductType` and `getActiveProvidersByProductType` are merged into one method, `getProvidersByProductType`, which now has a boolean flag parameter `onlyActive`.
9. `getProduct` is replaced by `getProviderWithProduct`, which returns `ProviderWithProduct`
10. `storagePricePerTibPerMonth` is now `storagePricePerTibPerDay`

#### Changes
* define and configure BloomSet with k=16
* move PDPOffering struct to a testing helper library
* redefine required schema as BloomSet16
* only enforce required schema probabilistically
* ensure synapse has a good method for fetching these keys all at once
* remove misleading `exists`
* remove ipni capability flags from required schema
* fix tests
* add and test BigEndian helper library for encoding and decoding integers

### DIFF
--- a/service_contracts/test/PDPOffering.sol
+++ b/service_contracts/test/PDPOffering.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.30;
 
+import {BigEndian} from "../src/lib/BigEndian.sol";
 import {ServiceProviderRegistry} from "../src/ServiceProviderRegistry.sol";
 import {ServiceProviderRegistryStorage} from "../src/ServiceProviderRegistryStorage.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
@@ -30,21 +31,21 @@ library PDPOffering {
             if (hash == keccak256("serviceURL")) {
                 schema.serviceURL = string(values[i]);
             } else if (hash == keccak256("minPieceSizeInBytes")) {
-                schema.minPieceSizeInBytes = abi.decode(values[i], (uint256));
+                schema.minPieceSizeInBytes = BigEndian.decode(values[i]);
             } else if (hash == keccak256("maxPieceSizeInBytes")) {
-                schema.maxPieceSizeInBytes = abi.decode(values[i], (uint256));
+                schema.maxPieceSizeInBytes = BigEndian.decode(values[i]);
             } else if (hash == keccak256("ipniPiece")) {
-                schema.ipniPiece = abi.decode(values[i], (bool));
+                schema.ipniPiece = BigEndian.decode(values[i]) > 0;
             } else if (hash == keccak256("ipniIpfs")) {
-                schema.ipniIpfs = abi.decode(values[i], (bool));
+                schema.ipniIpfs = BigEndian.decode(values[i]) > 0;
             } else if (hash == keccak256("storagePricePerTibPerDay")) {
-                schema.storagePricePerTibPerDay = abi.decode(values[i], (uint256));
+                schema.storagePricePerTibPerDay = BigEndian.decode(values[i]);
             } else if (hash == keccak256("minProvingPeriodInEpochs")) {
-                schema.minProvingPeriodInEpochs = abi.decode(values[i], (uint256));
+                schema.minProvingPeriodInEpochs = BigEndian.decode(values[i]);
             } else if (hash == keccak256("location")) {
                 schema.location = string(values[i]);
             } else if (hash == keccak256("paymentTokenAddress")) {
-                schema.paymentTokenAddress = abi.decode(values[i], (IERC20));
+                schema.paymentTokenAddress = IERC20(address(uint160(BigEndian.decode(values[i]))));
             }
         }
         return schema;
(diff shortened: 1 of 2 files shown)


service_contracts/src/ServiceProviderRegistry.sol
```diff
@@ -795,29 +739,21 @@ contract ServiceProviderRegistry is
...
-        require(bytes(pdpOffering.location).length > 0, "Location cannot be empty");
-        require(bytes(pdpOffering.location).length <= MAX_LOCATION_LENGTH, "Location too long");
+        uint256 foundKeys = 0;
+        for (uint256 i = 0; i < capabilityKeys.length; i++) {
+            uint256 key = BloomSet16.compressed(capabilityKeys[i]);
+            if (BloomSet16.mayContain(requiredKeys, key)) {
+                foundKeys |= key;
+            }
+        }
+        require(BloomSet16.mayContain(foundKeys, requiredKeys));
```
wjmelements: needs error to say missing key, though hard to say which key is missing

service_contracts/test/PDPOffering.sol
```diff
@@ -0,0 +1,101 @@
...
+        returns (Schema memory schema)
+    {
+        require(keys.length == values.length, "Keys and values arrays must have same length");
+        for (uint256 i = 0; i < keys.length; i++) {
+            bytes32 hash = keccak256(bytes(keys[i]));
+            if (hash == keccak256("serviceURL")) {
+                schema.serviceURL = string(values[i]);
+            } else if (hash == keccak256("minPieceSizeInBytes")) {
+                schema.minPieceSizeInBytes = abi.decode(values[i], (uint256));
+            }
```
wjmelements: should also use else everywhere

could binary search if this was in `src`.

wjmelements: 5a3224c

service_contracts/src/ServiceProviderRegistry.sol
```diff
@@ -467,37 +446,16 @@ contract ServiceProviderRegistry is
...
     /// @param productType The type of product to retrieve
-    /// @return productData The encoded product data
     /// @return capabilityKeys Array of capability keys
     /// @return isActive Whether the product is active
     function getProduct(uint256 providerId, ProductType productType)
         external
         view
         providerExists(providerId)
-        returns (bytes memory productData, string[] memory capabilityKeys, bool isActive)
+        returns (string[] memory capabilityKeys, bool isActive)
```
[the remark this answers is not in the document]
wjmelements: key existence is a signal

rvagg:
`getActiveProvidersByProductType` -> let's just make an `onlyActive` bool argument to `getProvidersByProductType` and ditch a method

service_contracts/src/ServiceProviderRegistry.sol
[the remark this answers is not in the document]
wjmelements: I think you still don't understand. If a capability is a boolean, its existence is sufficient. But that existence cannot be the signaled with the empty string because it is indistinguishable when doing a single key lookup in a solidity mapping.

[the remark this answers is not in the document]
rvagg: > If a key's absence unambiguously signals the capability is not supported, then the key's presence can unambiguously signal the capability is supported. Any nonzero length is thus truthy.

This is what I've been arguing for here and which is why I want an `exists` boolean return any time I want to ask for a specific key. I don't want to have to put a value in the value map, I just want to know the key exists and then not care about the value, and to work around the limitations of not having a `null` or non-zero sentinel in solidity. But we agree that the current implementation of doing that is broken - it should do it properly by iterating over keys that it has and figuring out whether it exists or not. But I also now think we can just do away with that entirely. There may be a case for "tell me the value for this key" or "tell me if you have this key", but with the way this is shaping up, I think all I really ever want out of this is be able to get the _full_ product, keys and values, and deal with it on the client side.

rvagg:
Trying out my own feedback as a PR: https://github.com/FilOzone/filecoin-services/pull/328

rvagg:
I'm just coming to terms with the per-day here vs per-month before and per-month that we have in FWSS, it's a bit odd that we have two versions of this. Now an SP has to divide the per month charge that everyone talks about to figure out how many days. We have a standard for what a "month" is in epochs that we use everywhere, it's not abnormal to encode a month as 30 days.

Anyway, just my 2c, not a big deal but it's jarring as I update Curio to work with this and think through what an SP has to deal with. The default I have to encode is 83333333333333333 to get close to the 2.5 USDFC we have in WarmStorage.

Original thread is https://github.com/FilOzone/filecoin-services/issues/297#issuecomment-3413851304

service_contracts/src/lib/BloomSet.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+pragma solidity ^0.8.30;
+
+/// @notice Bloom Filter with fixed params:
+/// @notice k = 16 bits per item
+/// @notice m = 256 bits per set
+/// @dev probability of false positives by number of items: (via https://hur.st/bloomfilter/?m=256&k=32)
```
Kubuxu:
```suggestion
/// @dev probability of false positives by number of items: (via https://hur.st/bloomfilter/?m=256&k=16)
```

wjmelements: ah yes I was originally doing 32 but 16 works better for slightly larger sets

PR #315, opened 2025-10-22
fix(712): use constant typehash instead of rehashing

rvagg:
just a minor whoopsie I noticed while cruising


wjmelements:
This might not actually change the code. I think solidity can evaluate the hash of string constants at compile time. Check the codesize before and after; might be the same.

Still a good change.

Issue #305, opened 2025-10-20
Non-sequential piece additions: how hard?


rvagg:
@wjmelements one for you to comment on.

There's some amount of complexity in the SDK right now to deal with the sequential nature of piece additions, we're doing a lot of state management, which I'm mostly fine with, but it's the sequential piece addition part of this that prevents an alternative interaction mode where everything is an atomic piece that can be bundled together in multiple ways. @hugomrdias is driving toward independent "actions" that bundle together to form the opinionated whole(s), but the actions themselves are ~atomic, and mostly independent (and stackable).

So, we do batched piece management in Synapse's StorageContext, mostly to deal with the problem of sequential nonces for pieces. PDPVerifier cares about order, because a data set is just a contiguous array of bytes that keeps on growing. There may be cases where a user cares about that order too, which is fine but that can be dealt with the multi-piece AddPieces() operation, or managing sequential addition locally. But as far as the common case user is concerned, they ought to be able to throw pieces at the contract in arbitrary order, multiple per tipset, have them land and find their natural order in PDPVerifier and _not_ have to be able to predict that order ahead of time to come up with the right `firstAdded` number for their 712 blob.

Can we do something similar with AddPieces to what we did with CreateDataSet? My hunch is that this is going to add more state because we're currently only tracking piece IDs for metadata, so we'd need something to track a new nonce field which we'd have to stick into `extraData`.

The main question is: how hard would this be? Then we need to figure out if it's worth doing.

wjmelements:
I think the main difficulty would be with the PDP sumtree. It would be less efficient if it were more sparse. I'm also concerned about the lingering state growth in FWSS.

PR #317, opened 2025-10-22
feat: use random nonces for AddPieces

rvagg:
Allows for non-sequential piece additions. PDPVerifier allocates its own pieceIds according to order added, we receive that as firstAdded and use it to refer to the same pieceIds as PDPVerifier. Now, instead of using firstAdded as a nonce, we allow for random nonces that the user now supplies separately in extraData.

* Adds a new "nonce" into extraData
* Repurposes clientDataSetIds as clientNonces
* Moves signature to the end of extraData to be consistent with CreateDataSet

Closes: https://github.com/FilOzone/filecoin-services/issues/305

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageServiceStateView.sol
+++ b/service_contracts/src/FilecoinWarmStorageServiceStateView.sol
@@ -22,12 +22,12 @@ contract FilecoinWarmStorageServiceStateView is IPDPProvingSchedule {
         return service.challengeWindow();
     }
 
-    function clientNonces(address payer, uint256 nonce) external view returns (uint256) {
-        return service.clientNonces(payer, nonce);
-    }
-
     function clientDataSets(address payer) external view returns (uint256[] memory dataSetIds) {
         return service.clientDataSets(payer);
+    }
+
+    function clientNonces(address payer, uint256 nonce) external view returns (uint256) {
+        return service.clientNonces(payer, nonce);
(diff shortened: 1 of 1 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -723,8 +723,15 @@ contract FilecoinWarmStorageService is
...
         // Decode the extra data
-        (bytes memory signature, string[][] memory metadataKeys, string[][] memory metadataValues) =
-            abi.decode(extraData, (bytes, string[][], string[][]));
+        (uint256 nonce, string[][] memory metadataKeys, string[][] memory metadataValues, bytes memory signature) =
+            abi.decode(extraData, (uint256, string[][], string[][], bytes));
+
+        // Validate nonce hasn't been used (replay protection)
+        require(clientNonces[payer][nonce] == 0, Errors.ClientDataSetAlreadyRegistered(nonce));
+        // Mark nonce as used and store dataset ID (lower 128 bits) and (firstAdded+1) (upper 128 bits)
+        // We add 1 to firstAdded to distinguish from CreateDataSet (which stores 0 in upper bits)
```
wjmelements: consider adding pieceData.length instead of 1, which would give it the meaning of cumulative piece count

wjmelements:
check the test failure

PR #320, opened 2025-10-22
feat: floor price set to the 0.06 USDFC/month

### LINKED ISSUE #319: Set up minimum price for ProofSet
We want to mitigate two scenarios

- Scenario 1: existing small datasets are not dropped by SPs when base fee raises;
- Scenario 2: a new datasets which get initiated with very small first piece(s) are not accepted by SP.

If basefee raises to 0.0056 nanoFIL (march 2025 avg)

We did some calculation

According to the [PDP gas and cost calculator ](https://lucaniz.github.io/pdp-gas-calculator/), datasets with
-  Size lower than 24 GiB
-  The max number of roots we considered so far  (~100k )

which translates into avg root size of 250Kb are not profitable for SPs in the current payment model where clients pay for actual storage, if basefee raises to 0.0056 nanoFIL (average value of march 2025). 

Given we are targeting clients with small dataset, this could represent an obstacle as small datasets can be rejected from storage in case basefee goes up.

In order to mitigate this potential issue, we propose to set up the minimum monthly payment for dataset to be 0.06 USDFC/month, whihc is the price a 24GB dataset would pay, assuming 2.5 USDFC/TiB/month.

Considering we are creating 2 copies per client, the total minimal cost on the client side would be 0.12 USDFC/month

Note that this min price would cover both Scenario 1 and Scenario 2.

Payment logic would stay the same, but until dataset size reaches 24 GiB, price would be constant and equal to 0.06USDFC/month.

Beyond that threshold, pricing can be the same as we have today, based on actual dataset size.

rvagg:
_Edit: This got switched to having 0.06 USDFC as the floor, not the bytes; so now the bytes works out as ~24.576 GiB and we get to talk about it in terms of cents per month._

(PR on top of #316, that should be merged first)

Amounts below this pay as if they are storing 24 GiB, which is 0.05859375 USDFC / month.

Closes: https://github.com/FilOzone/filecoin-services/issues/319

---

*  24 GiB = 24/1024 TiB = 0.0234375 TiB
* At 2.5 USDFC/TiB/month, price = 0.0234375 * 2.5 = 0.05859375 USDFC/month

Are we OK with 0.05859375? I have an option here of either setting 24GiB as the floor _size_ or 0.06 as the floor _price_.

If we go with 0.06 as the floor then we're working with 24.576 GiB floor size.

---

I also had to change the mock token from 6 decimal places to 18, cause there's already a floor rounding bit of code in here and it was getting stuck on a higher floor due to not enough decimal places, so everything came out wrong.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -587,6 +587,17 @@ contract FilecoinWarmStorageService is
 
         // Create the payment rails using the FilecoinPayV1 contract
         FilecoinPayV1 payments = FilecoinPayV1(paymentsContractAddress);
+
+        // Pre-check that payer has sufficient available funds to cover minimum storage rate
+        // This provides early feedback but doesn't guarantee funds will be available when SP calls nextProvingPeriod
+        (,, uint256 availableFunds,) = payments.getAccountInfoIfSettled(usdfcTokenAddress, createData.payer);
+        // Calculate required lockup: multiply first to preserve precision
+        uint256 minimumLockupRequired = (MINIMUM_STORAGE_RATE_PER_MONTH * DEFAULT_LOCKUP_PERIOD) / EPOCHS_PER_MONTH;
+        require(
+            availableFunds >= minimumLockupRequired,
+            Errors.InsufficientFundsForMinimumRate(createData.payer, minimumLockupRequired, availableFunds)
+        );
+
         uint256 pdpRailId = payments.createRail(
             usdfcTokenAddress, // token address
             createData.payer, // from (payer)
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -588,15 +588,8 @@ contract FilecoinWarmStorageService is
         // Create the payment rails using the FilecoinPayV1 contract
         FilecoinPayV1 payments = FilecoinPayV1(paymentsContractAddress);
 
-        // Pre-check that payer has sufficient available funds to cover minimum storage rate
-        // This provides early feedback but doesn't guarantee funds will be available when SP calls nextProvingPeriod
-        (,, uint256 availableFunds,) = payments.getAccountInfoIfSettled(usdfcTokenAddress, createData.payer);
-        // Calculate required lockup: multiply first to preserve precision
-        uint256 minimumLockupRequired = (MINIMUM_STORAGE_RATE_PER_MONTH * DEFAULT_LOCKUP_PERIOD) / EPOCHS_PER_MONTH;
-        require(
-            availableFunds >= minimumLockupRequired,
-            Errors.InsufficientFundsForMinimumRate(createData.payer, minimumLockupRequired, availableFunds)
-        );
+        // Validate payer has sufficient funds and operator approvals for minimum pricing
+        validatePayerOperatorApprovalAndFunds(payments, createData.payer);
 
         uint256 pdpRailId = payments.createRail(
             usdfcTokenAddress, // token address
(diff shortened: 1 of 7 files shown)


Kubuxu:
Is it feasible to add a check that the floor price rate is available to be paid when doing `createDataSet`? It would solve one of the problems causing: https://github.com/filecoin-project/curio/issues/726

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1200,18 +1203,26 @@ contract FilecoinWarmStorageService is
...
      * @param totalBytes Total size of the stored data in bytes
      * @return The storage rate per epoch
      */
     function _calculateStorageRate(uint256 totalBytes) internal view returns (uint256) {
-        return calculateStorageSizeBasedRatePerEpoch(totalBytes, STORAGE_PRICE_PER_TIB_PER_MONTH);
+        // Calculate natural size-based rate
+        uint256 naturalRate = calculateStorageSizeBasedRatePerEpoch(totalBytes, STORAGE_PRICE_PER_TIB_PER_MONTH);
+
+        // Calculate minimum rate (floor price converted to per-epoch)
+        uint256 minimumRate = MINIMUM_STORAGE_RATE_PER_MONTH / EPOCHS_PER_MONTH;
```
wjmelements: this might be zero. how many decimals are there? If only 6, this is zero.

wjmelements: I have checked; there are 18, so we are good
https://filfox.info/en/token/f410fqc4y2ovat777evodxjfcieir74jgf4cfcyah4wq?t=contract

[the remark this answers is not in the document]
rvagg: oh, thanks for the reminder, it's not supposed to be the only place, we need to expose it via `getServicePrice()` for the SDK, and we need it per month there and we can't multiply it out to reconstitute it because there's rounding problems with per-epoch (that we'll mostly ignore for now), so we should keep storing MINIMUM_STORAGE_RATE_PER_MONTH and I'll just add that accessor to ServicePricing

rvagg:
> Is it feasible to add a check that the floor price rate is available to be paid when doing createDataSet?

Oh, this is a really good idea, and yes it is possible! We have a `getAccountInfoIfSettled` on FilecoinPay that we can use to check their balance (if they were fully settled) so we can compare that with what would be needed for initial lockup and reject early. With create+add this is nice and clean. With create, then add, it's a little more messy and we lose guarantees but you'd just get rejected at add if you used up funds in between the two calls. If nextProvingPeriod gets done immediately after create+add then Curio should have a pretty solid guarantee that there is at least 0.06 on the table if they prove for a month, with a slight window for a possible race.

See latest commit in here for that feature.

Kubuxu:
Does that also cover `OperatorRateAllowanceExceeded`?

rvagg:
Figured out how to add additional checks in there. I'm of two minds about this, mostly it's great UX, we catch errors early and can be descriptive about it - good for SPs and good for users. But it also adds complexity and gets FWSS more deeply involved in something that it feels like should be something should be able to ask Filecoin Pay to figure out for us. But it's probably fine.

* `InsufficientFundsForMinimumRate` - not enough deposit
* `OperatorNotApproved` - need to approve FWSS
* `InsufficientRateAllowance` - not enough approval for the minimum rate (0.06/month)
* `InsufficientLockupAllowance` - not enough lockup approval for minimum rate @ 30 days
* `InsufficientMaxLockupPeriod` - not enough lockup period approval for the 30 day lockup

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
Kubuxu: It might get optimised out

Kubuxu:
> But it's probably fine.

We will be able to remove it when we combine `createDataSet` and initial `nextProvePeriod`.

Kubuxu (approved):
SGTM

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
Kubuxu: I'm almost certain you are right. Still, I'm ok paying that cost of division right now.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1177,21 +1234,29 @@ contract FilecoinWarmStorageService is
...
     /**
      * @notice Calculate storage rate per epoch based on total storage size
-     * @dev Returns storage rate per TiB per month
+     * @dev Returns storage rate per TiB per month with minimum pricing floor applied
      * @param totalBytes Total size of the stored data in bytes
      * @return storageRate The PDP storage rate per epoch
      */
-    function calculateRatesPerEpoch(uint256 totalBytes) external view returns (uint256 storageRate) {
-        storageRate = calculateStorageSizeBasedRatePerEpoch(totalBytes, STORAGE_PRICE_PER_TIB_PER_MONTH);
+    function calculateRatePerEpoch(uint256 totalBytes) external view returns (uint256 storageRate) {
```
wjmelements, line 1241: I think this method is unused. I plan to move it to View as part of the #469 refactors

wjmelements: https://github.com/FilOzone/filecoin-services/pull/476

PR #267, opened 2025-10-08
[WIP] : Update loop inside validatePayment to loop over proving periods (Issue #252)

parzival1821:
## Related Issue
Closes 252 ( Once completed)

## Current Progress
I've implemented the updated looping algorithm in validatePayment as described in the issue
- [x] Replaced the existing loop with new algorithm 
- [ ] Convert provenPeriods to a bitmap
- [ ] Remove provenThisPeriod

## What I Need Feedback On
**Before I continue with the rest of the issue, I'd like to get feedback on my implementation of the looping algorithm.**

Since there were no existing tests for this specific part of the code, I want to make sure my approach is correct before implementing the remaining parts of the issue. Also I would love to help you guys write some tests for the same if required

## Questions for Maintainers
- Does the new looping logic match what you had in mind?
- Are there any edge cases I should consider?
- Any suggestions for improvement before I move forward?

## Changes Made
- Modified `FilecoinWarmStorageService.sol` to implement the new loop structure

## Next Steps (after approval)
Once you confirm this approach is correct, I'll:
- Complete the remaining tasks in #252
- Request a full review

**Note:** This is a WIP/Draft PR specifically for early feedback on the algorithm implementation.
(Also, just a heads-up, I changed my username from `dev-n-dough` to `parzival1821`

PR #321, opened 2025-10-22
feat(IValidator): Allow Forward Progress Despite Missing Proofs

wjmelements:
Reviewer @rvagg
It is only important to block forward progress through the current proving period.
Currently we don't allow forward progress through unsettled periods without proofs.
This could be a problem if there was a rail unproven for a long time but proofs were later resumed.
Though much less likely after #258 and #267, we would not want it to be possible for the gap to exceed the block gas limit.

Also important to recall that the `fromEpoch` is an exclusive index while `toEpoch` is an inclusive index.
#### Changes
* allow to settleUpTo previous period when unproven
* define settledUpTo in previously irrelevant case

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -1465,7 +1465,7 @@ contract FilecoinWarmStorageService is
 
         uint256 startingPeriod = _provingPeriodForEpoch(activationEpoch, fromEpoch + 1);
 
-        // handle first period separately
+        // handle first period separately; it may be partially settled already
         uint256 startingPeriodDeadline = _calcPeriodDeadline(activationEpoch, startingPeriod);
 
         if (toEpoch < startingPeriodDeadline) {


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1443,26 +1443,22 @@ contract FilecoinWarmStorageService is
         }
 
         // Count proven epochs and find the last proven epoch
-        (uint256 provenEpochCount, uint256 lastProvenEpoch) =
+        (uint256 provenEpochCount, uint256 settledUpTo) =
```
rvagg: I feel like "settled" isn't the right language for (and in) _findProvenEpochs, maybe "settleUpTo" as an instruction would be more accurate? Or even "provenUpTo" as a description—although that's tricky because it's not necessarily all proven, it's more like "this is the last historical epoch within which you can make a settlement". The event itself has "settle" so I think it's just this internal usage that's making me trip over what on earth _findProvenEpochs is trying to do for us.

rvagg: k, well let's remove the "d" from here and in _findProvenEpochs, it's odd

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -910,15 +909,6 @@ contract FilecoinWarmStorageService is
             emit FaultRecord(dataSetId, faultPeriods, provingDeadlines[dataSetId]);
         }
 
-        // Record the status of the current/previous proving period that's ending
-        if (provingDeadlines[dataSetId] != NO_PROVING_DEADLINE && provenThisPeriod[dataSetId]) {
-            // Determine the period ID that just completed
-            uint256 completedPeriodId = getProvingPeriodForEpoch(dataSetId, provingDeadlines[dataSetId] - 1);
-
-            // Record whether this period was proven
-            provenPeriods[dataSetId][completedPeriodId >> 8] |= 1 << (completedPeriodId & 255);
```
wjmelements: this is already set by possessionProven

wjmelements: The code has changed over time. Before it was a bitmap, this could set both true and false. Since I also see things initialized to false elsewhere, perhaps the original author did not know that false was the default.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1442,78 +1434,71 @@ contract FilecoinWarmStorageService is
...
-        uint256 currentPeriod = getProvingPeriodForEpoch(dataSetId, block.number);
-
         if (fromEpoch < activationEpoch - 1) {
             fromEpoch = activationEpoch - 1;
         }
 
-        uint256 startingPeriod = getProvingPeriodForEpoch(dataSetId, fromEpoch + 1);
+        uint256 startingPeriod = _provingPeriodForEpoch(activationEpoch, fromEpoch + 1);
 
         // handle first period separately
```
rvagg:
```suggestion
        // handle first period separately to account for partial epoch count (fromEpoch > deadline)
```

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1442,78 +1434,71 @@ contract FilecoinWarmStorageService is
...
         uint256 startingPeriodDeadline = _calcPeriodDeadline(activationEpoch, startingPeriod);
 
         if (toEpoch < startingPeriodDeadline) {
-            if (_isPeriodProven(dataSetId, startingPeriod, currentPeriod)) {
+            if (_isPeriodProven(dataSetId, startingPeriod)) {
                 provenEpochCount = toEpoch - fromEpoch;
-                settledUpTo = toEpoch;
+                settleUpTo = toEpoch;
+            } else {
+                settleUpTo = fromEpoch;
```
rvagg: is this the main fix? it only applies to a small subset of possible outcomes in this method but I don't think I see anything else materially changing in this PR

this is a good fix, but how is it preventing forward progress now?

wjmelements:
> looks correct, I'm just not clear on what the fix is, `settleUpTo` seems to be `0` coming out of this currently if you choose a range shorter than one period and that period isn't proven

Before this change, settleUpTo is ignored in the case where no epochs are proven. After this change it is not.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: The main fix is https://github.com/FilOzone/filecoin-services/pull/321#discussion_r2459027847 but this change is also required, else `settleUpTo` can be `0` which Payments will then use, resetting the rail.

PR #325, opened 2025-10-23
feat(pricing): make pricing constants mutable with owner-controlled u…

rjan90:
Makes the three pricing rate constants mutable storage variables instead of immutable, enabling future price adjustments without requiring contract upgrades.

## Changes

- **Pricing Variables**: Converted `STORAGE_PRICE_PER_TIB_PER_MONTH`, `CACHE_MISS_PRICE_PER_TIB_PER_MONTH`, and `CDN_PRICE_PER_TIB_PER_MONTH` from immutable constants to mutable storage variables
- **New Function**: Added `updatePricing()` - owner-only function that allows selective updates to pricing rates (pass 0 to keep existing value)
- **New Getter**: Added `getCurrentPricingRates()` - public view function to query current pricing values
- **New Event**: Added `PricingUpdated` event to track pricing changes
- **Lockup Amounts**: `DEFAULT_CDN_LOCKUP_AMOUNT` and `DEFAULT_CACHE_MISS_LOCKUP_AMOUNT` remain immutable as they are not expected to change

## Impact
- New price rates will apply to all active data sets on their next proving period update (via `nextProvingPeriod()`)

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -1399,15 +1399,6 @@ contract FilecoinWarmStorageService is
     }
 
     /**
-     * @notice Get the current pricing rates
-     * @return storagePrice Current storage price per TiB per month
-     * @return minimumRate Current minimum monthly storage rate
-     */
-    function getCurrentPricingRates() external view returns (uint256 storagePrice, uint256 minimumRate) {
-        return (storagePricePerTibPerMonth, minimumStorageRatePerMonth);
-    }
-
-    /**
      * @notice Get the effective rates after commission for both service types
      * @return serviceFee Service fee (per TiB per month)
      * @return spPayment SP payment (per TiB per month)
--- a/service_contracts/abi/FilecoinWarmStorageService.abi.json
+++ b/service_contracts/abi/FilecoinWarmStorageService.abi.json
@@ -283,24 +283,6 @@
   },
   {
     "type": "function",
-    "name": "getCurrentPricingRates",
-    "inputs": [],
-    "outputs": [
-      {
-        "name": "storagePrice",
-        "type": "uint256",
-        "internalType": "uint256"
-      },
-      {
-        "name": "minimumRate",
-        "type": "uint256",
-        "internalType": "uint256"
-      }
-    ],
-    "stateMutability": "view"
-  },
-  {
-    "type": "function",
     "name": "getEffectiveRates",
     "inputs": [],
     "outputs": [
--- a/service_contracts/abi/FilecoinWarmStorageServiceStateLibrary.abi.json
+++ b/service_contracts/abi/FilecoinWarmStorageServiceStateLibrary.abi.json
@@ -297,6 +297,30 @@
   },
   {
     "type": "function",
+    "name": "getCurrentPricingRates",
+    "inputs": [
+      {
+        "name": "service",
+        "type": "FilecoinWarmStorageService",
+        "internalType": "contract FilecoinWarmStorageService"
+      }
+    ],
+    "outputs": [
+      {
+        "name": "storagePrice",
+        "type": "uint256",
+        "internalType": "uint256"
+      },
+      {
+        "name": "minimumRate",
+        "type": "uint256",
+        "internalType": "uint256"
+      }
+    ],
+    "stateMutability": "view"
+  },
+  {
+    "type": "function",
     "name": "getDataSet",
     "inputs": [
       {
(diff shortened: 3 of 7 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1316,20 +1344,34 @@ contract FilecoinWarmStorageService is
...
+     * @return storagePrice Current storage price per TiB per month
+     * @return cacheMissPrice Current cache miss price per TiB per month
+     * @return cdnPrice Current CDN price per TiB per month
+     */
+    function getCurrentPricingRates()
+        external
+        view
+        returns (uint256 storagePrice, uint256 cacheMissPrice, uint256 cdnPrice)
+    {
+        return (storagePricePerTibPerMonth, cacheMissPricePerTibPerMonth, cdnPricePerTibPerMonth);
```
wjmelements: I think these cdn rates should be zero after https://github.com/FilOzone/filecoin-services/pull/237. Might be an oversight in that PR. I have raised this question on slack.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -464,6 +466,32 @@ contract FilecoinWarmStorageService is
...
+     * @dev Only callable by the contract owner. Pass 0 to keep existing value unchanged.
+     * @param newStoragePrice New storage price per TiB per month (0 = no change)
+     * @param newCacheMissPrice New cache miss price per TiB per month (0 = no change)
+     * @param newCdnPrice New CDN price per TiB per month (0 = no change)
+     */
+    function updatePricing(uint256 newStoragePrice, uint256 newCacheMissPrice, uint256 newCdnPrice)
+        external
+        onlyOwner
+    {
+        require(newStoragePrice > 0 || newCacheMissPrice > 0 || newCdnPrice > 0, "At least one price must be non-zero");
```
wjmelements: we use abi `Errors` for this file to reduce codesize

service_contracts/src/Errors.sol
```diff
@@ -326,4 +326,13 @@ library Errors {
...
     );
+
+    /// @notice At least one price parameter must be non-zero when updating pricing
+    error AtLeastOnePriceMustBeNonZero();
+
+    /// @notice Price update exceeds the maximum allowed value
+    /// @param priceType The type of price being updated (e.g., "storage", "minimumRate")
+    /// @param maxAllowed The maximum allowed value for this price type
+    /// @param actual The attempted value that exceeds the maximum
+    error PriceExceedsMaximum(string priceType, uint256 maxAllowed, uint256 actual);
```
wjmelements: we want to avoid strings in the code because solidity strings are a lot of codesize. They are actually the reason we are using these ABI `error`s.

wjmelements: Alternatively you can use two different errors.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1341,22 +1389,31 @@ contract FilecoinWarmStorageService is
...
         });
     }
 
+    /**
+     * @notice Get the current pricing rates
+     * @return storagePrice Current storage price per TiB per month
+     * @return minimumRate Current minimum monthly storage rate
+     */
+    function getCurrentPricingRates() external view returns (uint256 storagePrice, uint256 minimumRate) {
+        return (storagePricePerTibPerMonth, minimumStorageRatePerMonth);
```
wjmelements: rm or move to WarmStorageView lib

wjmelements: https://github.com/FilOzone/filecoin-services/pull/330

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -467,6 +481,40 @@ contract FilecoinWarmStorageService is
...
+        }
+        if (newMinimumRate > 0) {
+            require(
+                newMinimumRate <= MAX_MINIMUM_STORAGE_RATE_PER_MONTH,
+                Errors.PriceExceedsMaximum(
+                    Errors.PriceType.MinimumRate, MAX_MINIMUM_STORAGE_RATE_PER_MONTH, newMinimumRate
+                )
+            );
+            minimumStorageRatePerMonth = newMinimumRate;
+        }
```
rvagg: I don't love having `if`s here and am not sure "Pass 0 to keep existing value unchanged." is all that helpful. It might help with footguns, perhaps, but now there's no way to set a floor of `0`, we'd have to use `1` to be our floor if we decided to undo that.
But, not a big deal if we're just trying to get this over the line at this point.

PR #341, opened 2025-11-05
Fault report spike

ZenGround0:
Not convinced we should commit this but here is a demo of how to interact with ethcalls to get a complete fault report.  State viewer has a few basic methods which make checking this information simple.

This is a claude heavy PR so some mistakes are going to be in here and the writing is a little AI-y.  But I've tested it out and it works pretty nicely.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,223 @@
+# Filecoin Services Scripts
+
+This directory contains utility scripts for interacting with Filecoin Warm Storage Service contracts.
+
+## Fault Status Query Scripts
+
+Two equivalent scripts are provided for querying fault status:
+- `query-fault-status.fish` - Fish shell version
+- `query-fault-status.sh` - Bash version (more widely compatible)
+
+Both scripts provide the same functionality to query and analyze the fault history of a dataset in the Filecoin Warm Storage Service.
+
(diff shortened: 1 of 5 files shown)


ZenGround0:
@rvagg I'm considering putting something like this into filecoin-pin datasets command.  Any thoughts?

scripts/README.md
```diff
@@ -0,0 +1,223 @@
...
+- Bit value 0 = period faulted (proof not submitted)
+
+#### Contract Queries
+
+The script makes the following RPC calls using `cast call`:
+
+1. `provingActivationEpoch(uint256)` - When proving started
+2. `provingDeadline(uint256)` - Current deadline block
+3. `getMaxProvingPeriod()` - Length of each proving period
+4. `challengeWindow()` - Size of the challenge window
```
rvagg: these two can and should be `getPDPConfig()` where they are both available on the returned struct; we've just been discussing removing these methods because they should have been cleaned up when we made it a single call to get everything.

BigLep:
@rvagg @rjan90 : I put this as M4.5 as I assume not needed for GA.  If this is a potential "good first issue", maybe we convert to an issue and let someone else take on?

PR #342, opened 2025-11-05
feat: Expand validatePayerOperatorApprovalAndFunds() to deal with CDN…

### LINKED ISSUE #339: Expand `validatePayerOperatorApprovalAndFunds()` to deal with CDN lockup requirements
If a data set has `withCDN` then the client will have two additional rails with fixed lockup amounts, one with `DEFAULT_CACHE_MISS_LOCKUP_AMOUNT` and the other with `DEFAULT_CDN_LOCKUP_AMOUNT`.

Unfortunately, `validatePayerOperatorApprovalAndFunds()` doesn't account for this in its nice "do you have enough" checks, and it probably should.

When doing this, we should make sure we're only calling `hasCDNMetadataKey(createData.metadataKeys)` once because it's not trivial, we can pass a boolean in to `validatePayerOperatorApprovalAndFunds()` and re-use it when we set up the rails.

DarkLord017:
fixes #339 
This pull request updates the `FilecoinWarmStorageService` contract to improve handling of CDN-related payment validation and updates the corresponding tests.

### DIFF
--- a/service_contracts/abi/Errors.abi.json
+++ b/service_contracts/abi/Errors.abi.json
@@ -273,27 +273,6 @@
   },
   {
     "type": "error",
-    "name": "InsufficientFundsForMinimumRate",
-    "inputs": [
-      {
-        "name": "payer",
-        "type": "address",
-        "internalType": "address"
-      },
-      {
-        "name": "minimumRequired",
-        "type": "uint256",
-        "internalType": "uint256"
-      },
-      {
-        "name": "available",
-        "type": "uint256",
-        "internalType": "uint256"
-      }
-    ]
-  },
-  {
-    "type": "error",
     "name": "InsufficientLockupAllowance",
     "inputs": [
       {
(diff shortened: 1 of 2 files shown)


rvagg (approved):
This is great, the only hesitation I have is the name of the error `InsufficientFundsForMinimumRate` which isn't quite right here, but I think we'll have to let that one pass since we're already working with that name.

I'd like @juliangruber || @bajtos to have a quick scan of this to verify that we're correct in adding the two default lockup amounts here to the minimum required funds to get a data set created if CDN.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1173,15 +1177,26 @@ contract FilecoinWarmStorageService is
...
+
+        uint256 totalMinimumLockup = minimumLockupRequired + extraCdnLockup;
+
         // Check that payer has sufficient available funds
         (,, uint256 availableFunds,) = payments.getAccountInfoIfSettled(usdfcTokenAddress, payer);
         require(
-            availableFunds >= minimumLockupRequired,
-            Errors.InsufficientFundsForMinimumRate(payer, minimumLockupRequired, availableFunds)
+            availableFunds >= totalMinimumLockup,
+            Errors.InsufficientFundsForMinimumRate(payer, totalMinimumLockup, availableFunds)
```
pyropy: I agree with @rvagg here, maybe the `Errors.InsufficientFundsForMinimumRate` is not exactly the best name, maybe something like `Errors.Errors.InsufficientLockupFunds` would be more accurate.

pyropy (commented):
Great job!

Let's maybe improve the error name before shipping this.

service_contracts/src/FilecoinWarmStorageService.sol
rvagg: I was suggesting we just put up with the name, but I believe if we deployed a new implementation contract this isn't going to break the proxy for us, it'll just yield a different error type and we'd need new ABI in Synapse to decode this error. But that's OK I think. We may as well fix it.

Issue #346, opened 2025-11-11
Document FWSS upgrade process


BigLep:
Tracking issue related to https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1762449236114639

wjmelements:
calibnet contracts are not verified 

going to try to verify the session key registry https://github.com/FilOzone/filecoin-services/releases/tag/v1.0.0#:~:text=0x97Dd879F5a97A8c761B94746d7F5cfF50AAd4452

wjmelements:
My [transaction](https://calibration.filfox.info/en/message/bafy2bzacedqvetrvcpmjvpy2ydkzvz6symcmu5m7ba7patkpojgc6xrx2yocu?t=4) has been stuck in the calibration mempool.

rjan90:
This issue has been kept open as a placeholder for documenting a FWSS upgrade process, since the steps has been documented by @wjmelements here. I have therefore tuned the titled appropriately.

PR #348, opened 2025-11-11
fix(scripts): deploy warm storage only

wjmelements:
Reviewer @rvagg @zenground0
Fixes a few things with the deploy-warm-storage-implementation-only script
#### Context
Toward https://github.com/FilOzone/filecoin-services/issues/346
#### Changes
* remove unused FILBEAM_CONTROLLER_ADDRESS
* use existing SIGNATURE_VERIFICATION_LIB_ADDRESS if provided
* fix `--libraries` arg
* use `-f 0x000000000000000000000000000000000000000000` with `cast call` to skip password prompt

### DIFF
--- a/service_contracts/tools/upgrade.sh
+++ b/service_contracts/tools/upgrade.sh
@@ -25,7 +25,7 @@ if [ -z "$PASSWORD" ]; then
 fi
 
 if [ -z "$CHAIN" ]; then
-  CHAIN=$(cast chain-id")
+  CHAIN=$(cast chain-id)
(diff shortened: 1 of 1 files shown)


wjmelements:
I also noticed another issue, in `tools/announcePlannedUpgrade.sh`, that `cast call`, when `ETH_KEYSTORE` is set, will prompt for a password. I don't think it needs the private key though so I will look for better fixes than adding `--password` to `cast call`.

rvagg (approved):
TIL about -f 0x000...

Issue #350, opened 2025-11-12
Modify rail rate at piecesAdded time, not nextProvingPeriod


rvagg:
Currently we have a window of time between when you add a piece and when we change the rate of the rail to pay for it, potentially up to the proving window size in time (24h on mainnet). This separation means that the user can add a piece, not have enough funds to pay for the new lockup amount by the time the rate change happens, leaving the SP with the decision on what to do with a piece they onboarded.

Instead, we should move the `updatePaymentRates()` call out of `nextProvingPeriod` and in to `piecesAdded`. Unfortunately we don't have the data set leaf count in `piecesAdded`, but we can call in to `PDPVerifier#getDataSetLeafCount(dataSetId)` to get it, this value is updated before `piecesAdded` is called so it should be accurate. This means that if the user doesn't have funds, their piece add should fail gas estimation.

rvagg:
When doing this we should document clearly the trade-off here and that we're slightly biasing toward the SP in this decision:

1. We are doing this to ensure that the client can pay the SP for the piece, so the SP doesn't hold onto the piece and only discover at `nextProvingPeriod` that the client can't, or won't pay for it.
2. By making this switch we are forcing the client to start paying for the piece _before_ any proofs are submitted to the chain, so the SP may technically not even have the piece and therefore fail their first proving window, but the rail will already have a new segment at a different rate which will fall within the previous proving window which may pass.

Additionally we need to be careful of **deletions**. We _schedule_ deletions that occur at `nextProvingPeriod`. One of the purposes of `nextProvingPeriod` is to consolidate a data set - additions and removals, so you end up with a new total size. Either rail modification for deletion should continue to happen in `nextProvingPeriod` (because it's always a reduction we don't have the same error condition we're solving for with piece addition), or we could even move rail modification to the schedulePieceDeletion call.

PR #352, opened 2025-11-14
feat(piecesAdded):validatePayerFunds

### LINKED ISSUE #351: validatePayerFunds for piecesAdded
We have `validatePayerOperatorApprovalAndFunds` for `dataSetCreated` with nice error types. We need the same for `piecesAdded` to validate that we have enough funds to cover the new lockup for the data set according to the leaf count of that data set. As per https://github.com/FilOzone/filecoin-services/issues/350 we don't have leaf count there, but we can get it from PDPVerifier and do the same calculation as is done in `updatePaymentRates` to work out the new rate and the new 30 day lockup amount from that and see that the user has funds, and if not, give a nice error.

DarkLord017:
fixes #351 

* Added a new internal function `validatePayerOperatorApprovalAndFundsForPieces` to `FilecoinWarmStorageService` that checks payer/operator approvals, available funds, and allowance limits before allowing new pieces to be added. 
* Imported the `IPDPVerifier` interface to enable querying the current leaf count from the PDP verifier, which is used in the new validation logic.
* Corrected mockpdpverifier to return leaf count

does the implementation loog good should i move to adding tests ?

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -664,7 +664,7 @@ contract FilecoinWarmStorageService is
 
         // Validate payer has sufficient funds and operator approvals for minimum pricing
         // If CDN is enabled, validation must account for the additional fixed lockup amounts
-        validatePayerOperatorApprovalAndFunds(payments, createData.payer, hasCDN , 0 , new Cids.Cid[](0));
+        validatePayerOperatorApprovalAndFunds(payments, createData.payer, hasCDN, 0, new Cids.Cid[](0));
 
         uint256 pdpRailId = payments.createRail(
             usdfcTokenAddress, // token address
@@ -840,7 +840,7 @@ contract FilecoinWarmStorageService is
         // to cover the increased per-epoch rate and the 30-day lockup implied by the
         // new total leaf count after adding these pieces.
         FilecoinPayV1 payments = FilecoinPayV1(paymentsContractAddress);
-        validatePayerOperatorApprovalAndFunds(payments, payer, false , dataSetId , pieceData);
+        validatePayerOperatorApprovalAndFunds(payments, payer, false, dataSetId, pieceData);
 
         // Store metadata for each new piece
         for (uint256 i = 0; i < pieceData.length; i++) {
--- a/service_contracts/test/FilecoinWarmStorageService.t.sol
+++ b/service_contracts/test/FilecoinWarmStorageService.t.sol
@@ -1068,7 +1068,7 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         );
 
         // Expected minimum: (0.06 USDFC * 86400) / 86400 = 0.06 USDFC = 6e16
-        uint256 minimumRequiredPerEpoch = (uint256(6e16) / (2880*30));
+        uint256 minimumRequiredPerEpoch = (uint256(6e16) / (2880 * 30));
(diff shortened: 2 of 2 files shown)


service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -1161,6 +1163,65 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         assertEq(dataSetId, 1, "Dataset should be created with above-minimum funds");
     }
 
+    // function testInsufficientFunds_AddPieces() public {
```
rvagg: what's going on here? why is it commented out?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -873,6 +881,57 @@ contract FilecoinWarmStorageService is
...
+        address payer,
+        uint256 dataSetId,
+        Cids.Cid[] memory pieceData
+    ) internal view {
+        // Get existing leaf count from the PDP verifier
+        uint256 leaves = IPDPVerifier(pdpVerifierAddress).getDataSetLeafCount(dataSetId);
+
+        uint256 totalBytes = leaves * BYTES_PER_LEAF;
+        uint256 storageRatePerEpoch = _calculateStorageRate(totalBytes);
+        uint256 lockupRequired = storageRatePerEpoch * DEFAULT_LOCKUP_PERIOD;
```
rvagg: this doesn't account for _new_ leaves being added, you'll need to work it out from the pieceData, something like:

```solidity
      uint256 addedLeaves = 0;
      for (uint256 i = 0; i < pieceData.length; i++) {
          (uint256 padding, uint8 height,) = Cids.validateCommPv2(pieceData[i]);
          addedLeaves += Cids.leafCount(padding, height);
      }
```

and add that to the current number of `leaves`

i.e. we're checking that they have funds for adding new data, not just for the existing data in the data set

rvagg: ah, correct you are! good work tracing that through

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -869,10 +877,63 @@ contract FilecoinWarmStorageService is
...
+            uint256 leaves = IPDPVerifier(pdpVerifierAddress).getDataSetLeafCount(dataSetId);
+            updatePaymentRates(dataSetId, leaves);
             emit PieceAdded(dataSetId, pieceId, pieceData[i], pieceKeys, pieceValues);
         }
     }
 
+    /// @notice Validate payer/operator approvals and funds for adding pieces
+    /// @dev Computes the new storage rate and corresponding 30-day lockup after adding `pieceData`
+    ///      and validates the payer has sufficient available funds and operator allowances.
+    function validatePayerOperatorApprovalAndFundsForPieces(
```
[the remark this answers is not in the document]
DarkLord017: okay

DarkLord017:
> Aside from the changes above, we should take a critical look at what happens when we do #350 because maybe all of this nice work is unnecessary and would be a waste of gas. Since we're doing a rail modification in the call anyway we're going to be getting these checks by Filecoin Pay. My lingering question is over the UX of it—do we get what we want in terms of error information from Filecoin Pay because I _think_ the only thing this might add after #350 is nicer sugar consistent with the CreateDataSet checks.

Please see #365

DarkLord017:
Also ig this  pr will resolve the issue #351 completely

PR #365, opened 2025-12-06
feat: modify rail rate when piecesAdded is called

DarkLord017:
Solves #350 

I added validatePayerFunds in piecesAdded instead of updating it in nextProvingPeriod.

When a piece is scheduled for deletion, we can’t remove it directly there because the payment needs to be updated. So I didn’t remove it from nextProvingPeriod, but added a check to see if pieces were actually removed.

The call will revert if the payment rates are not updated and nothing was deleted, so this check prevents that.


rvagg:
+136 bytes for this, 1,850 bytes margin remaining on contract size

Issue #336, opened 2025-10-30
Deprecate challengeWindow() function


rjan90:
## Description

As mentioned in [issue #335](https://github.com/FilOzone/filecoin-services/issues/335#issuecomment-3465223375), the `challengeWindow()` external view function should be deprecated once `challengeWindowSize` is made public.

## Current Situation

The `challengeWindow()` function is currently exposed as an external view function in:
- `service_contracts/src/FilecoinWarmStorageServiceStateView.sol` (line 21-23)

```solidity
function challengeWindow() external view returns (uint256) {
    return service.challengeWindow();
}
```

This function retrieves the challenge window size from storage using `extsload`.

## Background

In issue #335, it was decided to make `challengeWindowSize` public so that tools can fetch this information directly from the contract instead of hard-coding it. Once `challengeWindowSize` is made public, the `challengeWindow()` function becomes redundant and should be deprecated.

akronim26:
> > once challengeWindowSize is made public.
> 
> I don't think we should do this, it increases contract size and it's already available on `getPDPConfig` on the view contract. I think the only action here is to mark a `@deprecated` on the old function.

Thanks for the clarification. I’ve updated the implementation accordingly and opened PR #338, which only marks challengeWindow() as @deprecated. Please let me know if any further adjustments are needed.

akronim26:
@rvagg I have traced the issue through the codebase and it appears the root cause is that `maxProvingPeriod` and `challengeWindowSize` are currently `private` variables in `FilecoinWarmStorageService.sol`. Making these variables `public` would resolve this. Please let me know if this aligns with the intended fix.

rvagg:
Sorry @akronim26 I went ahead and did this myself cause I want to get some of these potentially-GA items queued up today: https://github.com/FilOzone/filecoin-services/pull/372

PR #372, opened 2025-12-19
fix!: remove redundant PDP config getters from view contract

rvagg:
Remove challengeWindow(), getChallengesPerProof(), and getMaxProvingPeriod() from external API. Use getPDPConfig() instead per IPDPProvingSchedule interface.

Closes: FilOzone/filecoin-services#336

---

This is breaking, there's not an easy path to deprecation so this just removes them - you have to use `getPDPConfig()` to get these going forward if you're not already. Curio uses this already. Synapse uses the old ones, for a `pieceStatus()` API which isn't core functionality and can be easily changed: https://github.com/FilOzone/synapse-sdk/issues/520

### DIFF
--- a/service_contracts/abi/FilecoinWarmStorageServiceStateLibrary.abi.json
+++ b/service_contracts/abi/FilecoinWarmStorageServiceStateLibrary.abi.json
@@ -1,25 +1,6 @@
 [
   {
     "type": "function",
-    "name": "challengeWindow",
-    "inputs": [
-      {
-        "name": "service",
-        "type": "FilecoinWarmStorageService",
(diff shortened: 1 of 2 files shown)


rvagg:
@rjan90 this change is for the view contract only, so when we deploy a new view contract with these changes we'd need to `setViewContract(newViewAddress)` on FWSS to activate this. Clients (Curio and Synapse) already know to navigate from FWSS to `viewContractAddress` to get access to all of this functionality. This can be independent of an FWSS upgrade too.

ZenGround0 (approved):
Thanks for the detailed description of where the breaks hit us

PR #373, opened 2026-01-02
feat: add devnet support to deploy

redpanda-f:
Make localnet be chain ID 1414, and use env vars


rvagg (approved):
rebased on main, removed submodule changes, all good to go now

PR #376, opened 2026-01-06
chore(docs): additional clarity and justification regarding rate change semantics

rvagg:


rvagg:
Added as second commit here after investigating https://github.com/filecoin-project/curio/issues/860 a bit deeper. The rate change semantics after termination are a bit subtle and I learnt some things during investigation. I've tried to capture that here.

The `RateChangeNotAllowedOnTerminatedRail` problem in https://github.com/filecoin-project/curio/issues/860 should already be handled by https://github.com/FilOzone/filecoin-services/pull/365 when we deploy it. Another (previously unknown) benefit of moving the rate change for AddPieces out of nextProvingPeriod!

rjan90 (approved):

Issue #375, opened 2026-01-02
Rail settlement check on Dataset Deletion


lordshashank:
I am thinking if we should have `rail.settledUpTo >= rail.endEpoch` (rail settlement) check in the delete dataset callback to fwss. Otherwise SP can bypass the validatePayment() check for rail settlement by sort of exploiting the settleTerminatedRailWithoutValidation() we have for saving client. Basically:-
- SP deletes dataset → breaks `validatePayment()` → normal `settleRail()` reverts
- Client's funds are locked (could be big amount depending on how big dataset is)
- Client is stuck in a dilemma:
  - Option A: Refuse to settle → Funds remain locked forever
  - Option B: Use `settleTerminatedRailWithoutValidation()` → Overpay the SP

Obvious tradeoff is to push SP to settle the dataset rail before dataset deletion, but this is what ideally should happen so I think doing this would save client from overpaying.

rvagg:
My attempt is here: #377

PR #377, opened 2026-01-07
fix(fwss)!: require rail settlement before dataset deletion

rvagg:
Block dataSetDeleted until settledUpTo >= endEpoch, preventing SPs from deleting datasets before clients can settle. Allow settlement to advance past unproven periods once their deadlines pass (with zero payment).

Closes: #375

### DIFF
--- a/service_contracts/abi/Errors.abi.json
+++ b/service_contracts/abi/Errors.abi.json
@@ -868,6 +868,27 @@
   },
   {
     "type": "error",
+    "name": "RailNotFullySettled",
+    "inputs": [
+      {
+        "name": "railId",
+        "type": "uint256",
+        "internalType": "uint256"
+      },
+      {
+        "name": "settledUpTo",
+        "type": "uint256",
+        "internalType": "uint256"
+      },
+      {
+        "name": "endEpoch",
+        "type": "uint256",
+        "internalType": "uint256"
+      }
+    ]
+  },
+  {
+    "type": "error",
     "name": "ServiceContractMustTerminateRail",
     "inputs": []
   },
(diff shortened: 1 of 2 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1612,7 +1626,7 @@ contract FilecoinWarmStorageService is
         (uint256 provenEpochCount, uint256 settleUpTo) =
             _findProvenEpochs(dataSetId, fromEpoch, toEpoch, activationEpoch);
 
-        // If no epochs are proven, we can't settle anything
+        // If no epochs are proven, no payment is due (but settlement may still advance)
```
ZenGround0: 🙏

ZenGround0 (approved):

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -4582,8 +4605,13 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         // Get updated info after termination
         info = viewContract.getDataSet(dataSetId);
 
-        // Wait for payment end epoch to elapse
-        vm.roll(info.pdpEndEpoch + 1);
```
wjmelements: Why do you remove this testcase?

rvagg: it's not removed, just extended because this is insufficient now, `settleRail()` must be called before deletion, see `vm.roll(endEpoch + maxProvingPeriod + 1)` below

PR #386, opened 2026-01-12
feat: ProviderIdSet

### LINKED ISSUE #382: Endorsed Providers Contract
**Context:** https://github.com/FilOzone/synapse-sdk/issues/540

### Summary

Create a simple contract that maintains a list of endorsed provider IDs. This replaces the endorsement certificates approach.

### Background

We have a three-tier provider trust model:

1. **All providers** - Everyone in ServiceProviderRegistry
2. **Approved providers** - Recorded in FilecoinWarmStorageService
3. **Endorsed providers** - Small set we store at least one copy with

This contract identifies the endorsed tier.

### Details

- Store a list of provider IDs
- Owner-controlled
- Suggested functions:
  - `getEndorsedProviders() → uint256[]` - returns full list
  - `addEndorsed(uint256 providerId)` - owner only
  - `removeEndorsed(uint256 providerId)` - owner only
  - Or just `setEndorsed(uint256[] providerIds)`?
- Deploy to calibnet and mainnet
- Add to deployment script(s) or make a new one (@rjan90 to provide input on preferred deployment script mechanism)
- Document contract address per network
- Later: Transfer to multisig and test with multisig before GA

### Follow-on Synapse work

- Add contract address constant per network, add override address in `SynapseOptions` for devnet use and other testing
- Add memoised query in `StorageManger`, fetch once per session, cache result, pass to dependent components
- Update https://github.com/FilOzone/synapse-sdk/pull/443 to use contract data instead of certs

_Note: part of the `upload()` work will be redoing a bunch of this code so the main priority is just to make something provably work._

wjmelements:
Closes #382
Reviewer @rvagg
ProviderIdSet is an Ownable arrayset of provider ids.
It uses less space than a bitset if the selected providerIds are sparse.
It uses 1/16 as much space as an [OpenZeppelin EnumerableSet](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol).
If we want to reuse this data structure in another smart contract, it can be converted into a library, but I do not do that here.
#### Interface
* `getProviderIds`
* `containsProviderId`
* `addProviderId`
* `removeProviderId`
#### Changes
* add ProviderIdSet
* test ProviderIdSet
* add deployment script
* deploy on mainnet and testnet


rvagg (approved):
Looks good. The expected use in the foreseeable future is 1-4 entries for this contract. At that scale the storage savings are minimal and a plain `uint64[]` would have similar cost with less code to maintain.

rvagg:
@TippyFlitsUK should we treat both calibnet SPs as equal quality so both "endorsed"? or is one a backup?

service_contracts/src/ProviderIdSet.sol
BigLep: @wjmelements : I came to this smart contract as a smart contract newbie thinking this would be an easy one to look at and learn from.  

In looking at and studying it, I'm wondering why we went with this level complexity for a temporary thing that will have small data storage and low churn?

My concerns:
- Takes extra time out of reviewers to understand and verify correctness
- While this by itself is not a big deal, I don't think it's setting the right precedent.  We're spending extra time to go with "endorsement contracts" to simplify things operationally but then it seems like we're adding complexity in the code itself.  

No doubt there is good learning and education for myself and others in this work, but maybe we do that as a gist or a ProviderIdSet-Advaced.sol file?

PR #332, opened 2025-10-27
build: prep FWSS v1.0.0 release

rjan90:
Prep General Availability (GA) release of the Filecoin Warm Storage Service contracts.

- Bump PDP submodule to v3.1.0 tag
- Bump FWSS and service provider registry version to v1.0.0
- Prep changelog

PR #388, opened 2026-01-13
chore!: consistent contract naming, add SignatureVerificationLib

rvagg:
BREAKING change because it renames keys here but they are very inconsistent and confusing and the PDP ones don't even show you what the proxy is.

---

I had to go use this today and was a bit frustrated it was less useful than I imagined.

@rjan90 I got `SIGNATURE_VERIFICATION_LIB_ADDRESS` off a message in Slack from you for calibnet but had to go hunting on the chain for the mainnet one, I'm pretty sure that's it by the bytecode match but you might just want to see if you have a record of that being the right address for mainnet.

### DIFF
--- a/service_contracts/tools/README.md
+++ b/service_contracts/tools/README.md
@@ -61,8 +61,8 @@ The `deployments.json` file stores deployment addresses organized by chain ID:
 {
   "314": {
     "PDP_VERIFIER_PROXY_ADDRESS": "0x...",
-    "PAYMENTS_CONTRACT_ADDRESS": "0x...",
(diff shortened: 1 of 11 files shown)


rjan90:
Confirming that both the `SIGNATURE_VERIFICATION_LIB_ADDRESS` address for:
- Calibnet is correct: https://github.com/FilOzone/filecoin-services/pull/332#issuecomment-3450847034
- Mainnet is correct: https://github.com/FilOzone/filecoin-services/pull/332#issuecomment-3478027044

How useful are these addresses? Is there a chance that these will be updated/changed?

wjmelements (approved):

PR #385, opened 2026-01-12
fix(fwss): ignore errors when CDN rails already terminated via FilecoinPay

pyropy:
Checks if payment rails are already terminated prior to the termination attempt.  

Closes #383

Issue #383, opened 2026-01-12
`terminateService` and `terminateCDNService` revert if CDN rails already terminated externally


rvagg:
Found by external auditor (Composable Security).

## Summary

CDN payment rails are intentionally created without a validator to support FilBeam's usage-based payment model (see #237). This allows the payer to terminate CDN rails directly via FilecoinPay. However, `terminateService` and `terminateCDNService` don't handle this case, causing them to revert when CDN rails are already terminated.

## Problem

When CDN rails are created, `validator` is intentionally set to `address(0)`: https://github.com/FilOzone/filecoin-services/blob/7b87a784a9ada2051bb0ef13bd28b3cc1ab8d31e/service_contracts/src/FilecoinWarmStorageService.sol#L694-L701

This means:

1. Payer can call `FilecoinPayV1.terminateRail(cacheMissRailId)` directly
2. No callback to FWSS (validator is `address(0)`)
3. FWSS is unaware the rails were terminated

Later, when `terminateService` is called: https://github.com/FilOzone/filecoin-services/blob/7b87a784a9ada2051bb0ef13bd28b3cc1ab8d31e/service_contracts/src/FilecoinWarmStorageService.sol#L1077-L1081

`payments.terminateRail(info.pdpRailId)`  succeeds but `payments.terminateRail(info.cacheMissRailId)` *reverts* with `RailAlreadyTerminated`, rolling back the entire transaction including the PDP rail termination.

Same issue exists in `terminateCDNService`: https://github.com/FilOzone/filecoin-services/blob/7b87a784a9ada2051bb0ef13bd28b3cc1ab8d31e/service_contracts/src/FilecoinWarmStorageService.sol#L1175-L1176

## Impact

- **Service termination blocked**: Neither payer nor service provider can terminate storage service through FWSS if CDN rails were terminated externally
- **CDN termination blocked**: FilBeam controller cannot call `terminateCDNService`

Note: `topUpCDNPaymentRails` is NOT affected since it already queries rail state from FilecoinPay and has appropriate error handling (lines 1142-1146).

## Proposed Solution

Check if CDN rails are already terminated before calling `terminateRail`:

**In `terminateService`:**

```solidity
if (deleteCDNMetadataKey(dataSetMetadataKeys[dataSetId])) {
    FilecoinPayV1.RailView memory cacheMissRail = payments.getRail(info.cacheMissRailId);
[...]

PR #390, opened 2026-01-14
chore(docs): document nuances of CDN interations

rvagg:
Ref: https://github.com/FilOzone/filecoin-services/pull/237
Ref: https://github.com/FilOzone/filecoin-services/issues/383
Ref: https://github.com/FilOzone/filecoin-services/pull/385

---

This depends on https://github.com/FilOzone/filecoin-services/pull/385 landing, but has other useful information.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -142,7 +142,12 @@ require(settledUpTo >= endEpoch, RailNotFullySettled)
 
 ## CDN Payment Rails
 
-Datasets with CDN support have three payment rails: a **PDP rail** for storage proving, and two **CDN rails** (cache-miss and bandwidth) for content delivery. CDN rails pay to the FilBeam beneficiary address, which is immutably set at contract deployment.
+Datasets with CDN support have three payment rails: a **PDP rail** for storage proving, and two **CDN rails** for content delivery:
+
+- **Cache-miss rail** (`cacheMissRailId`): Pays to the storage provider (SP) for origin fetches
+- **Bandwidth rail** (`cdnRailId`): Pays to the FilBeam beneficiary address (immutably set at deployment)
+
+Both CDN rails have `paymentRate = 0` and use fixed lockup for one-time payments based on usage.
 
 ### Payment Models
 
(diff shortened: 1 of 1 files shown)


rjan90 (approved):

SPEC.md
```diff
@@ -139,3 +139,33 @@ require(settledUpTo >= endEpoch, RailNotFullySettled)
...
+
+**PDP rail**: Uses proof-based settlement. FWSS acts as validator, receiving callbacks to verify that storage proofs were submitted before authorizing payment. Settlement amounts depend on proving status.
+
+**CDN rails**: Use usage-based settlement. No validator is set. The FilBeam controller calculates payment amounts based on actual egress metrics and calls `settleFilBeamPaymentRails()` on FWSS. This decouples CDN payment logic from the proof-based model used for storage.
+
+### External Rail Management
+
+CDN rails can also be managed directly through FilecoinPay without FWSS involvement:
+
+- **Top-up**: Clients call `topUpCDNPaymentRails()` on FWSS to increase their CDN lockup, which extends their egress allowance.
```
juliangruber: this is contradicting previous sentence, stating that it can be managed without FWSS involvement

SPEC.md
```diff
@@ -139,3 +139,33 @@ require(settledUpTo >= endEpoch, RailNotFullySettled)
...
+- **Settlement**: Anyone can call FilecoinPay's `settleRail()` on CDN rails since they have no validator. The FilBeam controller uses this for usage-based settlement.
```
juliangruber: https://github.com/FilOzone/filecoin-services/blob/3f132815997a0f6fc7b207e511b7d54205044838/service_contracts/src/FilecoinWarmStorageService.sol#L1092-L1118

rvagg:
@pyropy @juliangruber check latest commit.

The sync issue with `withCDN` is real but how much does it matter? It's mostly something to note for clients inspecting this I think.

If being able to call `terminateRail()` directly with FilecoinPay is a problem, we could add a validator to these rails. If the validator is FWSS then you could still let them call that but we could handle `railTerminated()` here which would allow clean-up and we wouldn't have the `withCDN` sync issue. `settleRail()` would still be a noop but we'd let that pass through here. So perhaps you want to rethink the no-validator design in light of this?

Issue #401, opened 2026-01-27
`immutable` reinitializer parameter


wjmelements:
To avoid future mistakes where we forget to update the initializer count, we can generate it at deployment time after fetching the previous value with `eth_getStorageAt`.

PR #402, opened 2026-01-27
feat: immutable reinitializer sequence number

wjmelements:
Closes https://github.com/FilOzone/filecoin-services/issues/401
Reviewers @zenground0 @rvagg
CC @rjan90
#### Context
Currently we have to increment this reinitializer parameter (called `version` in Initializable.sol, though we use semver for our `VERSION`) each time.
It would be better if we initialized it to the correct value at deployment time.
That way, we would not be able to mistakenly forget to increment it.
Immutable fields are set in the code during the constructor (initcode) rather than in storage.
#### Changes
* set immutable reinitializer parameter in constructor 
* update scripts that construct these implementations to supply the new constructor parameter

### DIFF
--- a/service_contracts/tools/get-initialized-counter.sh
+++ b/service_contracts/tools/get-initialized-counter.sh
@@ -1,4 +1,11 @@
 #!/bin/bash
+#
+# Reads the current reinitializer counter from an OpenZeppelin
+# Initializable proxy contract.
+#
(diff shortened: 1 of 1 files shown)


wjmelements:
The bool can only be true inside of a transaction, so it can never be read as true.

PR #406, opened 2026-01-30
fix(subgraph): use deployments.json as source of truth for addresses

nijoe1:
- Refactor generate-constants.js and generate-config.js to read contract
 - addresses from service_contracts/deployments.json instead of the manual
 - config/network.json. This ensures subgraph configuration stays in sync when contracts are redeployed.
 
 cc: @BigLep @rjan90 @rvagg


rvagg:
We're also pulling out the subgraph retriever from the sdk here since we've as yet had no signal of its use and it's awkward to maintain (cause we don't have a testing path for it and none of us use it): https://github.com/FilOzone/synapse-sdk/pull/575
Could be put back in, but would need a proper owner who cared about it.

PR #423, opened 2026-02-10
refactor(IValidator): simplify findProvenEpochs logic

ZenGround0:
This is follow up from reviewing #419.  The main difference is that the special cases are all handled together in each iteration of the loop.  I believe compact code like this easier to maintain though perhaps this is closer to a style thing than a law of nature.  I prefer dealing with a single gadget that takes a while to understand but then applies to all cases in a very dumb way (in this case iterating through all periods).  As opposed to many simpler gadgets that are put together in a more complex configuration.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -1684,30 +1684,22 @@ contract FilecoinWarmStorageService is
     /// payment, allowing the rail to eventually be fully settled and finalised even if the
     /// provider missed proofs.
     ///
-    /// The range (fromEpoch, toEpoch] may span part of a period, a full period, or
-    /// multiple periods. The function splits this into up to three regions:
+    /// Iterates through each proving period that overlaps the range (fromEpoch, toEpoch].
+    /// Partial periods at the start and end are handled by clamping each period's contribution
+    /// to [max(periodStart, fromEpoch), min(toEpoch, deadline)].
     ///
-    ///   1. First period (may be partial): fromEpoch may be mid-period (assume that prior epochs
-    ///      are already settled, likely due to a rail rate change, see below).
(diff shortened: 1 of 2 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1717,57 +1717,22 @@ contract FilecoinWarmStorageService is
...
-                settleUpTo = fromEpoch;
-            } else {
+        uint256 endingPeriod = _provingPeriodForEpoch(activationEpoch, toEpoch);
+        for (uint256 period = startingPeriod; period <= endingPeriod; period++) {
+            uint256 deadline = _calcPeriodDeadline(activationEpoch, period);
+            if (_isPeriodProven(dataSetId, period)) {
+                uint256 settleStart = max(deadline - maxProvingPeriod, fromEpoch);
+                uint256 settleEnd = min(deadline, toEpoch);
+                provenEpochCount += settleEnd - settleStart;
+                settleUpTo = min(toEpoch, deadline);
```
wjmelements: this is `settleEnd`

wjmelements:
This PR reduces codesize but adds redundant loads:
* FWSS codesize: -210
* `testValidatePayment_AllEpochsProven` gas: +373
* `testValidatePayment_SomeEpochsProven` gas: +427
* `testValidatePayment_NoEpochsProven` gas: +1482
* `testValidatePayment_MixedProvenAndUnprovenWithPassedDeadlines` gas: +351
* `testValidatePayment_FirstPeriodUnproven` gas: +2988

We're accepting this regression in exchange for improvements in maintainability and codesize.

> Redundant (warm) loads are only 100 gas in ethereum but I think they're not so cheap in Filecoin, where the storage cost increases as contract storage accumulates.

Some of these regressions are addressed in #424, without complicating the code, though the perf remains worse than before.

PR #424, opened 2026-02-10
perf(IValidator): reduce `maxProvingPeriod` loads

wjmelements:
Reviewer @rvagg @zenground0
These are simple changes that reduce the storage loads of #423 without making the code harder to understand.
After these changes, there is no benefit to making a local copy of maxProvingPeriod because solc does that optimization automatically.
#### Changes
* pure `provingPeriodForEpoch`
* `deadline += provingPeriodLength;`

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -131,20 +131,7 @@ Where base rail rate changes have occurred (e.g. pieces were added mid-period, c
 
 ### Settlement Algorithm (_findProvenEpochs)
 
-The function splits the settlement range `(fromEpoch, toEpoch]` into up to three regions:
-
-```
-fromEpoch      startingDeadline                  toEpoch  endingDeadline
-    |                 |                              |          |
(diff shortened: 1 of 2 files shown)


wjmelements:
[`deadline += maxProvingPeriod`](https://github.com/FilOzone/filecoin-services/pull/424/commits/6b958b88a1b444ca190536137373eb3d142266dc) makes the code slightly larger but also reduces loads

* FWSS codesize: +29
* `testValidatePayment_AllEpochsProven` gas: -128
* `testValidatePayment_NoEpochsProven` gas: -393
* `testValidatePayment_MixedProvenAndUnprovenWithPassedDeadlines` gas: -144
* `testValidatePayment_SomeEpochsProven` gas: -136

Any improvement greater than 100 is a reduction in loads. Redundant (warm) loads are only `100` gas in ethereum but I think they're not so cheap in Filecoin, where the storage cost increases as contract storage accumulates.

Issue #435, opened 2026-03-03
Options for the Sybil Fee Payment


timfong888:
Regarding: https://github.com/FilOzone/filecoin-services/issues/371

### Context                                                                                                                                                                               
                                                                  
  The SDK's default path already uses `createDataSetAndAddPieces` — one off-chain EIP-712 signature, no on-chain transaction from the client. The standalone `createDataSet` endpoint was   left open for empty-bucket UX but nobody uses it in production. 

  In go-synapse, the client calls `CreateDataSet` on-chain directly (`msg.sender` = client), so the client pays the 0.1 FIL sybil fee. This is a fundamentally different architecture — the client needs a funded FIL wallet and sends their own transaction.

To keep our current gasless UX while shifting the fee to the client, there are three paths:                                                                                               
                                                                  
  | Path | Additional client tx? | How it works |                                                                                                                                           
  |------|----------------------|-------------|                   
  | **Deduct from client's USDFC lockup** | No | Contract upgrade — FWSS deducts sybil fee equivalent from client's first payment/escrow. Client still only signs. |
  | **Client sends tx directly** | Yes | Breaks the gasless model. UX regression. go-synapse already does this — client is `msg.sender`, pays 0.1 FIL directly. |
  | **Increase floor price** | No | Bump minimum monthly charge so SP recovers sybil fee from storage revenue. Indirect — SP still out-of-pocket day one. Recovery period depends on FIL/USDFC exchange rate. |

wjmelements:
> The only remaining concern is if funds from a rail setup to flow to the SP's account can be redirected to the burnt funds actor.

I don't think so. We don't normally burn USDFC. Can't we pay it to the SP? It would compensate them for create rail gas.

jennijuju:
curious if there is any way to implement A without adding a manual list? i,e any magical hook that the PDP listener can implement that will send a call back to the pdp verifier to indicate when a dataset is created a sybil fee is burnt?

jennijuju:
can you shed some light on why  C cant be as simple as "always burn X from the payment rail" where X has no dependency on the value of sybil fee; my argument is that there exists an X that is high enough for the attack to be not rational.

ZenGround0:
This entire class of attacks is irrational.  There is never any expected reward from running this attack.  The attacker's goal is to harm SP or network by incurring annoying cost on them.  

If we aren't concerned about this class of things we should label this as wontfix.

rvagg:
[...]

PR #437, opened 2026-03-04
feat: dataset create client burn

ZenGround0:
This is the FWSS part of the solution arrived at in #435.

The rail operations are straightforward if a bit verbose and expensive.  We need the settle to actually finalize the rail since unfortunately modifyPayment doesn't do finalization of terminated rails.  

This was heavily AI driven.  The code itself is nicely cleaned up.  The tests might still be a little verbose or copy paste.  I'll see if I can improve them on my own but would appreciate any tips for doing so in review too.

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -200,7 +200,6 @@ contract FilecoinWarmStorageService is
     uint256 private constant TIB_IN_BYTES = GIB_IN_BYTES * 1024; // 1 TiB in bytes
     uint256 private constant EPOCHS_PER_MONTH = 2880 * 30;
 
-
     // Metadata size and count limits
     uint256 private constant MAX_KEY_LENGTH = 32;
     uint256 private constant MAX_VALUE_LENGTH = 128;
--- a/service_contracts/abi/FilecoinWarmStorageService.abi.json
+++ b/service_contracts/abi/FilecoinWarmStorageService.abi.json
@@ -55,19 +55,6 @@
   },
   {
     "type": "function",
-    "name": "USDFC_SYBIL_FEE",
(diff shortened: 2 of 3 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -692,6 +695,28 @@ contract FilecoinWarmStorageService is
...
+        // --- Burn rail: extract USDFC sybil fee from client and send to PDPVerifier ---
+        {
+            // Move funds to this contract in filecoin pay
+            uint256 burnRailId = payments.createRail(
+                usdfcTokenAddress,
+                createData.payer, // from: client
+                address(this), // to: FWSS collects and forwards
+                address(0), // no validator
+                0,
+                address(this)
```
wjmelements:
```suggestion
                usdfcTokenAddress,
                createData.payer, // from: client
                address(payments), // to: FilecoinPay auction
                address(0), // no validator
                0,
                address(0)
```

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -5407,6 +5365,130 @@ contract FilecoinWarmStorageServiceUpgradeTest is Test {
...
+        vm.stopPrank();
+    }
+
+    function _encodeCreateData(address payer, uint256 clientDataSetId) internal pure returns (bytes memory) {
+        string[] memory keys = new string[](0);
+        string[] memory values = new string[](0);
+        return abi.encode(payer, clientDataSetId, keys, values, FAKE_SIGNATURE);
+    }
+
+    function testDataSetCreation_SybilFeeBurned() public {
```
wjmelements: add a check in here that `payments.getRail` returns an empty rail. We don't want to leak storage on this otp rail.

PR #438, opened 2026-03-06
feat: add workflow for FWSS approved SP management

rjan90:
This PR adds convenient GHA tools for managing FWSS approved storage providers which are meeting the dealbot-criterias.

Changes included:
- add `service_contracts/tools/warm-storage-manage-approved-provider.sh`
  - supports `list`, `status`, `add`, and `remove`
  - resolves `providerId` from service provider address
  - auto-resolves the live removal index
  - can emit Safe calldata or propose directly to Filecoin Safe tx-service
- add `.github/workflows/manage-approved-provider.yml`
  - manual workflow for approved SP add/remove operations
  - supports `dry_run=true` for validation and `dry_run=false` for proposal submission
  - inputs are limited to `network`, `operation`, `provider_id` / `service_provider_address`, optional `safe_nonce`, and `dry_run`
  
  The runbook itself is in: https://www.notion.so/filecoindev/FOC-Operational-Excellence-2b7dc41950c1802aa432fff8ecb801cc?source=copy_link#31bdc41950c180c486deef64503debd0

### DIFF
--- a/.github/workflows/manage-approved-provider.yml
+++ b/.github/workflows/manage-approved-provider.yml
@@ -61,6 +61,20 @@ jobs:
         with:
           version: v1.3.5
 
+      - name: Input summary
+        run: |
+          {
+            echo "## Manage Approved Provider"
+            echo ""
+            echo "| Input | Value |"
+            echo "|---|---|"
+            echo "| Network | ${{ inputs.network }} |"
+            echo "| Operation | ${{ inputs.operation }} |"
+            echo "| Provider ID Input | ${{ inputs.provider_id || 'n/a' }} |"
+            echo "| Service Provider Address Input | ${{ inputs.service_provider_address || 'n/a' }} |"
+            echo "| Dry Run | ${{ inputs.dry_run }} |"
+          } >> "$GITHUB_STEP_SUMMARY"
+
       - name: Validate and optionally propose transaction
         env:
           ETH_RPC_URL: ${{ inputs.network == 'Calibnet' && secrets.CALIBNET_RPC_URL || secrets.MAINNET_RPC_URL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/manage-approved-provider.yml
```diff
@@ -0,0 +1,92 @@
...
+          SAFE_NONCE: ${{ inputs.safe_nonce }}
+          DRY_RUN: ${{ inputs.dry_run }}
+          PROPOSE_TO_SAFE: true
+          SAFE_PROPOSER_PRIVATE_KEY: ${{ secrets.SAFE_PROPOSER_PRIVATE_KEY }}
+          SAFE_TRANSACTION_SERVICE_API_KEY: ${{ secrets.SAFE_TRANSACTION_SERVICE_API_KEY }}
+        run: |
+          chmod +x service_contracts/tools/warm-storage-manage-approved-provider.sh
+          ./service_contracts/tools/warm-storage-manage-approved-provider.sh
+
+      - name: Summary
```
BigLep: Does it make sense to display this before running"Validate and optionally propose transaction"?  That way someone see what the script is going to be executed with.

Seeing it at the end may give the impression that we successfully ran with all those inputs which isn't necessarily true.

BigLep (approved):
Great to get this standardized and thanks for adding the runbook - nice!

I'm a bit worried about all the bash we're building up in this repo and I'm wondering with how many scripts we have in here if we would be benefiting to a toolchain/language that enables more modularity and code sharing.  But that's a bigger question that this PR adding one more script.

One thing to consider is to get some tools in to help make sure our shell scripts stay orderly (e.g., https://github.com/patrickvane/shfmt, https://github.com/koalaman/shellcheck - I recently learned about these from @redpanda-f ).  Anyways, that's also not a focus/concern for the scope of this PR, but was thinking about it while seeing the Bash mountain grow.

PR #450, opened 2026-03-20
ci: enforce additive-only storage layout changes

Chaitu-Tatipamula:
Resolves #446

### Description
Introduces a robust CI check to guarantee that updates to the UUPS proxy implement additive-only changes to [FilecoinWarmStorageServiceLayout.sol](cci:7://file:///Volumes/DISK1/PLDG/Filoz/filecoin-services/service_contracts/src/lib/FilecoinWarmStorageServiceLayout.sol:0:0-0:0). This strictly enforces that existing storage slots cannot be removed, reordered, or modified, effectively preventing storage collision bugs upon contract upgrades.

### Changes
* **Scripted Verification**: Added [check_storage_layout.sh](cci:7://file:///Volumes/DISK1/PLDG/Filoz/filecoin-services/service_contracts/tools/check_storage_layout.sh:0:0-0:0) which compares the actively generated storage slots against `origin/$GITHUB_BASE_REF` (or `HEAD~1`) to automatically detect dropped slots, shifted numbers, or illegal mid-contract insertions.
* **Workflow Integration**: Integrated into [.github/workflows/check.yml](cci:7://file:///Volumes/DISK1/PLDG/Filoz/filecoin-services/.github/workflows/check.yml:0:0-0:0) to automatically reject Pull Requests containing destructive layout edits prior to merge.
* **Developer Tooling**: Exposed the structural check via `make check-layout` to facilitate local testing prior to commit.
* **Documentation**: Updated [SPEC.md](cci:7://file:///Volumes/DISK1/PLDG/Filoz/filecoin-services/SPEC.md:0:0-0:0) reflecting the new "Done Criteria" requirements regarding automated verification and release-manager approvals.


BigLep (commented):
I'd like @rjan90 or a developer to give the final approval here.

rjan90:
One gap in this check is that it only compares `NAME SLOT` pairs from the generated layout constants:

- [extract_slots in `check_storage_layout.sh` (lines 20-25)](https://github.com/FilOzone/filecoin-services/blob/518ad2795c8cfafc6c9dd697e8a3f4a6d0e32d75/service_contracts/tools/check_storage_layout.sh#L20-L25)
- [generated `FilecoinWarmStorageServiceLayout.sol` output](https://github.com/FilOzone/filecoin-services/blob/518ad2795c8cfafc6c9dd697e8a3f4a6d0e32d75/service_contracts/src/lib/FilecoinWarmStorageServiceLayout.sol#L8-L30)

Because that generated file does not include type or offset metadata, a destructive storage change that keeps the same slot number, such as changing a field width/type or repacking values within the same slot, would still pass this check even though the proxy layout is no longer compatible.

If we want this to be a stronger upgrade-safety guard, we likely need to diff the full `forge inspect ... storageLayout` metadata, not just the generated slot constants.

service_contracts/Makefile
```diff
@@ -34,6 +35,10 @@ build:
 $(LAYOUT): tools/generate_storage_layout.sh src/FilecoinWarmStorageService.sol
 	$^ | forge fmt -r - > $@
 
+# Storage layout JSON (full metadata for upgrade safety checks)
+$(LAYOUT_JSON): src/FilecoinWarmStorageService.sol
+	forge inspect --json $< storageLayout | jq '[.storage[] | {label, slot, offset, type}]' > $@
+
```
rjan90, line 42: The new JSON snapshot is a nice improvement, but I think storing raw `.storage[].type` here may cause false positives. Forge includes compiler-generated IDs in user-defined types (for example` t_struct(Foo)6_storage`), and those IDs can change after harmless source edits even when the storage layout is unchanged. That would make the CI check fail on a spurious “type changed” diff. For a follow-up, it may be safer to canonicalize the type metadata or resolve it through `.types` before comparing.

Chaitu-Tatipamula: oh yes thanks for pointing this, i'll update the MakeFile to strip out those IDs before it writes to the JSON snapshot so it just leaves the pure type behind

rjan90: Hmmm... I now think we introduced  a more important false negative. For example, if `struct PlannedUpgrade`or `struct DataSetInfo` changes internally, the stored type string can r`emain struct FilecoinWarmStorageService.PlannedUpgrade` / `mapping(uint256 => struct FilecoinWarmStorageService.DataSetInfo)`, so the JSON does not change and the check still passes.

Instead of snapshotting only the top-level type label, recursively resolve the referenced `.types` entries and compare a normalized nested representation of the full type shape. If that is too much for this PR, we can at least narrow the docs/claims so we don’t say this catches all type changes yet.

rjan90: I pushed a small follow-up commit to close the remaining storage-layout gap above^^.

The previous canonicalization fixed the compiler-generated type ID false positive, but it only preserved the top-level type label. That meant a nested struct change could still pass if the outer storage entry stayed as something like struct PlannedUpgrade or mapping(uint256 => struct DataSetInfo).

The new commit keeps the stable top-level type string, but also adds a recursive typeDetails snapshot generated from Forge’s .types dictionary. The checker now compares typeDetails when available, so nested struct/member layout changes are caught while harmless compiler-ID churn still does not fail CI.

rjan90 (approved):

service_contracts/Makefile
Chaitu-Tatipamula: Thanks @rjan90

PR #464, opened 2026-04-29
chore: validate EIP-55 checksums in deployments.json, document PDPVerifier sync

rvagg:
Ref: https://github.com/FilOzone/synapse-sdk/pull/759
Ref: https://github.com/FilOzone/filecoin-services/pull/447

Surfaced in Synapse 759, the casing of PDP_VERIFIER_IMPLEMENTATION_ADDRESS got squashed, which breaks the checksum, which upsets some tooling (like Viem) that perform checks by default. Enforcing EIP-55 format is good practice regardless so this introduces a CI check for it.

Also update actions/checkout to latest.


service_contracts/tools/check_deployments_checksums.sh
```diff
@@ -0,0 +1,31 @@
...
+
+JSON_PATH="${1:-service_contracts/deployments.json}"
+ZERO="0x0000000000000000000000000000000000000000"
+FAIL=0
+N=0
+
+while IFS=$'\t' read -r path value; do
+    [[ "$value" =~ ^0x[0-9a-fA-F]{40}$ ]] || continue
+    [ "${value,,}" = "$ZERO" ] && continue
+    N=$((N + 1))
```
rjan90, line 16: Optional nit here, to make it also pass on default macOS /bin/bash 3.2.

```suggestion
    [[ "$value" =~ ^0x[0-9a-fA-F]{40}$ ]] || continue
    lower_value="$(printf '%s' "$value" | tr '[:upper:]' '[:lower:]')"
    [ "${value,,}" = "$ZERO" ] && continue
    N=$((N + 1))
```

rjan90 (approved):
Lgtm, one optional nit

.github/workflows/check.yml
```diff
@@ -83,3 +83,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: v1.3.5
+
```
rvagg:
```suggestion
```

service_contracts/tools/check_deployments_checksums.sh
rvagg: where's `lower_value` being used after this tho?

rvagg: ahhhh it's the `"${value,,}"` that's the bash 4 incompat, I can just make that go away entirely actually, the zero address is already a valid checksum so we don't even need that check!

hugomrdias (approved):
LGTM

PR #471, opened 2026-05-07
fix(fwss): price PDP rails on raw bytes, not Fr32-expanded leaf size

rvagg:
`updatePaymentRates` was multiplying leafCount by 32 to derive total bytes, which is the Fr32-expanded size, not the raw payload size. Switch to the new `Cids.leafCountToRawSize` to apply the 127/128 ratio, removing the ~0.787% per-byte overcharge.

Also marks `getDataSetSizeInBytes` deprecated and to-be-removed; it now delegates to the same helper rather than computing Fr32 size. Drops the `BYTES_PER_LEAF` constant.

Bumps lib/pdp submodule to FilOzone/pdp#266.

Fixes: #451

~_Pending: we're going to tag + release pdp with that change, but not deploy it, so I'll update here with a new head commit when that's done but it'll be the same code._~

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -22,7 +22,6 @@ import {Extsload} from "./Extsload.sol";
 import {SignatureVerificationLib} from "./lib/SignatureVerificationLib.sol";
 
 uint256 constant NO_PROVING_DEADLINE = 0;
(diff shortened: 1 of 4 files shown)


rvagg:
Submodule updated to https://github.com/FilOzone/pdp/releases/tag/v3.3.0

Kubuxu (approved):

ZenGround0 (approved):
LGTM

PR #472, opened 2026-05-07
fix(deploy): pass new PDPVerifier constructor args in warm-storage-deploy-all.sh

rvagg:
Came up in https://github.com/FilOzone/foc-devnet/pull/107


rjan90 (approved):
Nice!

Issue #469, opened 2026-05-01
Per-operation pricing for FilecoinWarmStorageService


rvagg:
## Project Tracking
This is the master tracking issue for the per-operation pricing changes for GA. Below is intended to make sure we're thinking wholistically on delivering for this effort:
- [x] Termination clarification / simplification
  - [x] https://github.com/FilOzone/filecoin-services/issues/477
  - [x] https://github.com/FilOzone/pdp/issues/269
  - [x] https://github.com/FilOzone/pdp/issues/271
  - [x] https://github.com/FilOzone/filecoin-services/issues/484
  - [x] https://github.com/filecoin-project/curio/issues/1235
  - [x] https://github.com/filecoin-project/curio/issues/1236
  - [x] https://github.com/filecoin-project/curio/issues/1237
  - [x] https://github.com/FilOzone/synapse-sdk/issues/792
  - [x] https://github.com/FilOzone/dealbot/issues/575
- [x] Client changes: 
  - [x] https://github.com/FilOzone/synapse-sdk/issues/763
  - [x] https://github.com/filecoin-project/filecoin-pin/issues/402
- [x] FWSS changes
  - [x] https://github.com/FilOzone/filecoin-services/issues/457
  - [x] https://github.com/FilOzone/filecoin-services/pull/482
  - [x] Per Operation Payments: https://github.com/FilOzone/filecoin-services/pull/488
  - [x] https://github.com/FilOzone/filecoin-services/issues/466
  - [x] https://github.com/FilOzone/filecoin-services/issues/496
  - [x] https://github.com/FilOzone/filecoin-services/pull/501
- [x] https://github.com/FilOzone/foc-devnet/issues/113
- [x] https://github.com/FilOzone/filecoin-services/issues/453
- [x] Public communication
  - 2026-07-15: further communication is being tracked in https://app.notion.com/p/filecoindev/Pricing-change-communication-391dc41950c18155924afe84d9baebd1?v=27b3e0656ba7455eb42a3d6341578e20
  - [x] https://github.com/FilOzone/synapse-sdk/issues/817
  - [x] https://github.com/FilOzone/filecoin-services/issues/468
  - [ ] filecoin.cloud updates?
  - [ ] email outreach with subscribers / known users?

## About

FWSS currently charges clients a single streaming USDFC rate per dataset, scaled by stored data with a minimum floor. Storage providers are expected to absorb the gas cost of every onchain operation.

[...]

PR #474, opened 2026-05-11
feat(fwss): emit dataset deletion auth signer

### LINKED ISSUE #457: Emit event from dataSetDeleted method, handle optional signed user auth
Extension of https://github.com/filecoin-project/curio/issues/1028#issuecomment-4174709026

`dataSetDeleted` in FWSS currently does _not_ receive an `extraData` because the SP does not need permission to delete. The proposal on the table is allow the user to call a delete API on Curio with a 712 signed auth, Curio can validate that auth by verifying the client's signature _or_ checking for active session keys for the client and verifying against one of them. If the client is authenticated, the SP can go ahead and perform a deletion on chain _and_ pass the 712 through the delete call which makes it to FWSS. In FWSS we'll perform the same verification, and then emit an event for deletion and the event will have a `signer` field that will either be `address(0)` if there was no `extraData` or the `extraData` didn't contain a valid 712 for the client, _or_ it will contain the signer address if it is valid (the signer address is either the client's address _or_ the session key address if it was signed by a session key).

Then we end up with a chain of evidence—why was it deleted, did the user authorise it or did the SP do it unilaterally.

rvagg:
Lift the DeleteDataSet EIP-712 typehash into SignatureVerificationLib and use it to optionally verify deletion auth passed through PDP extraData.

Emit an FWSS DataSetDeleted event with the authorized signer when the client or a valid session key signed the delete request, or address(0) when deletion is unauthenticated, malformed, or unilateral.

Closes: https://github.com/FilOzone/filecoin-services/issues/457


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1637,6 +1650,54 @@ contract FilecoinWarmStorageService is
...
+        bytes calldata signature = decodeDeleteDataSetSignature(extraData);
+        if (signature.length == 0) {
+            return address(0);
+        }
+
+        bytes32 digest =
+            _hashTypedDataV4(keccak256(abi.encode(SignatureVerificationLib.DELETE_DATA_SET_TYPEHASH, dataSetId)));
+
+        return SignatureVerificationLib.authorizedDeleteDataSetSigner(payer, signature, digest, sessionKeyRegistry);
+    }
```
Kubuxu, line 1675: Should we move this to SignatureVerificationLib to not increase the size of this contract?

wjmelements: Unfortunately, moving `_hashTypedDataV4` out of the contract is nontrivial. We would have to find another library.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1637,6 +1650,54 @@ contract FilecoinWarmStorageService is
...
+
+    /**
+     * @notice Decodes delete extraData without reverting on malformed optional data.
```
wjmelements: Perhaps we should revert on malformed data. If the data is malformed, probably the developer has a bug, and this would swallow that error rather than reporting it to them.

wjmelements: Similarly we could consider reverting if the authorization is expired.

rvagg: yeah, you're right; avoid footguns for the SP, I'll fix
I fully expect the SP to properly validate this before submitting so it'll be hard for it to be incorrect, but expired expirations would be one easy way for it to fail if you were close and the message landed late

rvagg: done: `InvalidDeleteDataSetExtraData` now if it's obviously invalid `extraData` and the normal `InvalidSignature` if the signature doesn't validate or authorisation is expired

service_contracts/src/FilecoinWarmStorageService.sol
rvagg: Latest commit saves 2 bytes, but not really by moving stuff.

We _could_ move the typehash creation out of here into the library, we'd want to do the same with schedulePieceRemovals, but then we incur a cross-contract call gas just to do a trivial `abi.encode` in both cases. The createDataSet and addPieces typehash creation is in the library right now, so conceptually everything belongs in there, but in those two cases there's a non-trivial amount of work to do.

This whole change so far adds 43 bytes to the contract. We still have 594 bytes headroom so the [pricing changes](https://github.com/FilOzone/filecoin-services/issues/469) may be forced to use library calls.

rvagg:
Converting to draft for now, https://github.com/FilOzone/filecoin-services/issues/469#issuecomment-4426175032 is an important point, the timing of using the 712 for this call makes an awkward gap from client->deletion. We need a better mechanism here to incentivise clean-up, make it so the client doesn't have to submit an on-chain msg under normal circumstances, and get the SP some funds to justify the clean-up work. Will move some design discussion back to https://github.com/FilOzone/filecoin-services/issues/457 first cause there are options.

rvagg:
Closing this pending https://github.com/FilOzone/filecoin-services/issues/477; same code could be reused but some of it will need to change.

BigLep:
@rvagg : I'm closing this given your comment in https://github.com/FilOzone/filecoin-services/pull/474#issuecomment-4457168506 . Please correct if that's wrong.

PR #476, opened 2026-05-13
feat!: Rails library

wjmelements:
Reviewer @rvagg
Toward #469
I am cutting a PR here in order to help keep the code review digestible.
I think reviewing the changes commit-by-commit might also be helpful.
This reduces contract size from `23939` to `19857`
#### Changes
* move pricing constants and functions to lib/PriceListUSDFC.sol
* move rails logic to lib/Rails.sol 
* move `calculateRatePerEpoch` to the View contract, update gen to allow pure methods in the view contract.
* remove methods `updateServiceCommission` and `updatePricing`; these parameters will now be managed by the upgrade path
* update deployment scripts to deploy the rails lib

### DIFF
--- a/service_contracts/abi/FilecoinWarmStorageService.abi.json
+++ b/service_contracts/abi/FilecoinWarmStorageService.abi.json
@@ -106,25 +106,6 @@
   },
   {
     "type": "function",
-    "name": "calculateRatePerEpoch",
-    "inputs": [
-      {
-        "name": "totalBytes",
-        "type": "uint256",
-        "internalType": "uint256"
-      }
-    ],
-    "outputs": [
-      {
-        "name": "storageRate",
-        "type": "uint256",
-        "internalType": "uint256"
-      }
-    ],
-    "stateMutability": "view"
-  },
-  {
-    "type": "function",
     "name": "configureProvingPeriod",
     "inputs": [
       {
@@ -302,7 +283,7 @@
         "internalType": "uint256"
       }
     ],
-    "stateMutability": "view"
+    "stateMutability": "pure"
   },
   {
     "type": "function",
@@ -800,37 +781,6 @@
   },
   {
     "type": "function",
-    "name": "updatePricing",
-    "inputs": [
-      {
-        "name": "newStoragePrice",
-        "type": "uint256",
-        "internalType": "uint256"
-      },
-      {
-        "name": "newMinimumRate",
-        "type": "uint256",
-        "internalType": "uint256"
-      }
-    ],
-    "outputs": [],
-    "stateMutability": "nonpayable"
-  },
-  {
-    "type": "function",
-    "name": "updateServiceCommission",
-    "inputs": [
-      {
-        "name": "newCommissionBps",
-        "type": "uint256",
-        "internalType": "uint256"
-      }
-    ],
-    "outputs": [],
-    "stateMutability": "nonpayable"
-  },
-  {
-    "type": "function",
     "name": "upgradeToAndCall",
     "inputs": [
       {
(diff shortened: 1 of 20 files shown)


service_contracts/src/lib/PriceListUSDFC.sol
```diff
@@ -0,0 +1,51 @@
...
+uint256 constant MIB_IN_BYTES = 1024 * 1024; // 1 MiB in bytes
+uint256 constant GIB_IN_BYTES = MIB_IN_BYTES * 1024; // 1 GiB in bytes
+uint256 constant TIB_IN_BYTES = GIB_IN_BYTES * 1024; // 1 TiB in bytes
+
+uint256 constant TOKEN_DECIMALS = 18;
+uint256 constant DEFAULT_LOCKUP_PERIOD = 2880 * 30; // 1 month (30 days) in epochs
+uint256 constant EPOCHS_PER_MONTH = 2880 * 30;
+
+// USDFC has 18 decimals, so $1 = 10**18 (a.k.a. ether)
+uint256 constant SYBIL_FEE = 0.1 ether;
```
rvagg, line 15: this was originally a const then we exposed it on PDPVerifier and used `IPDPVerifier(pdpVerifierAddress).USDFC_SYBIL_FEE()` and now we're back to a const here, is this intentional?

service_contracts/src/lib/Rails.sol
```diff
@@ -0,0 +1,243 @@
...
+pragma solidity ^0.8.20;
+
+import {Errors} from "../Errors.sol";
+import {FilecoinPayV1} from "@fws-payments/FilecoinPayV1.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {
+    DEFAULT_LOCKUP_PERIOD,
+    DEFAULT_CACHE_MISS_LOCKUP_AMOUNT,
+    DEFAULT_CDN_LOCKUP_AMOUNT,
+    DEFAULT_LOCKUP_PERIOD,
```
rvagg, line 10:
```suggestion
```

service_contracts/abi/FilecoinWarmStorageService.abi.json
```diff
@@ -104,25 +104,6 @@
     "outputs": [],
     "stateMutability": "nonpayable"
   },
-  {
-    "type": "function",
-    "name": "calculateRatePerEpoch",
```
rvagg, line 109: worth calling this out clearly in the docs / commit msg, we don't use it thankfully but someone _might_

rvagg:
SPEC.md needs updating, it still has _"Only the contract owner can update pricing by calling updatePricing(newStoragePrice, newMinimumRate) ..."_; that'll need fixing.

rvagg:
Oh, and README.md also says _"The service uses static global pricing set by the contract owner (default: 2.5 USDFC per TiB/month)"_. I guess it's still mostly true, just not a setter, we could leave this, your call.

rvagg (approved):
I normally hate per-commit reviewing but this is really nicely separated out, so it was actually a good experience (and I hadn't caught up on GitHub making it a bit easier too); very nice work.
Just a few things to take care of but this is a preemptive approval assuming you'll look at & take care of these things:
* the sybil fee fetcher question
* duplicate `DEFAULT_LOCKUP_PERIOD` import
* would like somewhere mentioning the calculateRatePerEpoch removal, this is technically a breaking change, we could signal it with `!` but calling it out in a commit message would be very helpful too, the commit `refactor: Rails.updateStorageRates` could become `refactor!: Rails.updateStorageRates` and in the body mention that `calculateRatePerEpoch` was moved off the proxy.
* SPEC.md needs updating to talk about how the pricing gets changed now

rvagg:
I think we could probably remove calculateRatePerEpoch entirely when we're done here, may not be the worst thing to remove it entirely even from here and just document the breakage

service_contracts/src/lib/PriceListUSDFC.sol
wjmelements: yeah, I think a const in the PriceList is the right way to go. We would run into trouble if the pdp verifier disagreed with this but maybe we can remove the pdp verifier's sybil check once we adopt the cleanup rebate system.

service_contracts/src/lib/Rails.sol
wjmelements: I wouldn't expect duplicate imports to pass lint and `fmt --check`

bae29ad

wjmelements:
> I think we could probably remove calculateRatePerEpoch entirely when we're done here, may not be the worst thing to remove it entirely even from here and just document the breakage

I'm inclined to keep it for now because it's used to test the pure function. It can be calculated from the information in `getServicePrice`, and isn't that useful on-chain, so I agree it should be removed as part of #469

PR #485, opened 2026-05-22
feat: termination extra data param

### LINKED ISSUE #457: Emit event from dataSetDeleted method, handle optional signed user auth

wjmelements:
Reviewer @rvagg
CC @LexLuthr
Toward https://github.com/FilOzone/filecoin-services/issues/469
Replaces https://github.com/FilOzone/filecoin-services/pull/474
This adds the next signature for terminateService which will allow optional extra data.
Cutting this separately while I work on the more complicated operation pricing change in order to unblock curio and synapse
#### Changes
* add terminateService method with extra data
* deprecate terminateService without extra data
* use extradata for client auth signature

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -103,6 +103,13 @@ contract FilecoinWarmStorageService is
         uint256 indexed dataSetId, uint256 indexed pieceId, Cids.Cid pieceCid, string[] keys, string[] values
     );
 
+    /// @notice Emitted when a service is terminated.
+    /// @param approver The address that authorized termination: the payer, one of the payer's
+    ///   session keys (SessionKeyRegistry), or the service provider. Cross-reference with
+    ///   `DataSetCreated` to classify: `approver == serviceProvider` is provider-initiated;
+    ///   otherwise the payer (or their session key) authorized it. Mutual termination — payer
+    ///   signed off-chain while the provider submitted the tx — is indistinguishable from
+    ///   payer-initiated using this event alone; inspect the call trace to detect it.
     event ServiceTerminated(
         address indexed approver,
         uint256 indexed dataSetId,
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -53,6 +53,12 @@ uint256 constant MAX_ADD_PIECES_EXTRA_DATA_SIZE = 8192; // 8 KiB
 */
 uint256 constant MAX_SCHEDULE_PIECE_REMOVALS_EXTRA_DATA_SIZE = 256; // 256 bytes
 
+/*
+* Maximum extraData for terminateService
+* Supports: signature (160 bytes needed)
+*/
+uint256 constant MAX_TERMINATE_SERVICE_EXTRA_DATA_SIZE = 256; // 256 bytes
+
 /// @title FilecoinWarmStorageService
 /// @notice An implementation of PDP Listener with payment integration.
 /// @dev This contract extends SimplePDPService by adding payment functionality
@@ -98,7 +104,11 @@ contract FilecoinWarmStorageService is
     );
 
     event ServiceTerminated(
-        address indexed caller, uint256 indexed dataSetId, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId
+        address indexed approver,
+        uint256 indexed dataSetId,
+        uint256 pdpRailId,
+        uint256 cacheMissRailId,
+        uint256 cdnRailId
     );
 
     event PDPPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId);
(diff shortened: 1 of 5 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -925,28 +935,46 @@ contract FilecoinWarmStorageService is
...
+            bytes memory signature = abi.decode(extraData, (bytes));
+            _verifyTerminateServiceSignature(info.payer, dataSetId, signature);
+            approver = info.payer;
+            // TODO if msg.sender is info.serviceProvider, termination is immediate
+        } else {
+            require(
+                msg.sender == info.payer || msg.sender == info.serviceProvider,
+                Errors.CallerNotPayerOrPayee(dataSetId, info.payer, info.serviceProvider, msg.sender)
+            );
+            approver = msg.sender;
```
rvagg, line 973: The `approver` field flips between signer and submitter depending on the path, so a subgraph or other indexer has to compare it against `transaction.from` just to know which case it's looking at.

A cleaner version would be to make the field one thing only, the `ecrecover`ed signer or `address(0)` when there's no `extraData`, and let `transaction.from` carry the submitter. That also gives us the session key identity for free (we currently emit `info.payer` and lose it), and a subgraph can't `ecrecover` calldata (I think, or at least it would be tricky) so emitting it is the only way it ever sees which key signed.

* `address(0)` - unilateral, actor is `transaction.from` (payer or SP, since the unsigned path is gated to those)
* non-zero - client authorised by that signer (resolve to payer via the registry or assume from fwss data set payer); mutual if `transaction.from == SP`, otherwise client pulling out via a delegate/relay (maybe a kind of MSP that was given a session key?)

Note this drops "non-zero = mutual" in the original design doc because you're not gating `extraData` being submittable only by teh SP, so that signal doesn't survive an open submit path (e.g. client delegates a session key to an MSP who terminates unilaterally). Mutual is really two questions: client consent and SP consent, so it can't live in one address, but it's still recoverable from the field plus `transaction.from`.

wjmelements: ~[7a16a4a](https://github.com/FilOzone/filecoin-services/pull/485/commits/7a16a4a4ae35ec891c8baac6fda601890312adee)~ (https://github.com/FilOzone/filecoin-services/pull/485#discussion_r3300368519)

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -925,28 +931,45 @@ contract FilecoinWarmStorageService is
...
+            bytes memory signature = abi.decode(extraData, (bytes));
+            signer = _verifyTerminateServiceSignature(info.payer, dataSetId, signature);
+            // TODO if msg.sender is info.serviceProvider, termination is immediate
+        } else {
+            require(
+                msg.sender == info.payer || msg.sender == info.serviceProvider,
+                Errors.CallerNotPayerOrPayee(dataSetId, info.payer, info.serviceProvider, msg.sender)
+            );
+            signer = address(0);
+        }
```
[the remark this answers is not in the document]
wjmelements: > With a valid payer/session-key signature, any relayer can terminate the dataset, which is a behavioral change from the previous authorization model.

Good.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -53,6 +53,12 @@ uint256 constant MAX_ADD_PIECES_EXTRA_DATA_SIZE = 8192; // 8 KiB
 */
 uint256 constant MAX_SCHEDULE_PIECE_REMOVALS_EXTRA_DATA_SIZE = 256; // 256 bytes
 
+/*
+* Maximum extraData for terminateService
+* Supports: signature (160 bytes needed)
+*/
+uint256 constant MAX_TERMINATE_SERVICE_EXTRA_DATA_SIZE = 256; // 256 bytes
+
```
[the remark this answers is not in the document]
wjmelements: This matches MAX_SCHEDULE_PIECE_REMOVALS_EXTRA_DATA_SIZE

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -925,28 +931,45 @@ contract FilecoinWarmStorageService is
...
+    function terminateService(uint256 dataSetId, bytes calldata extraData) external {
+        _terminateService(dataSetId, extraData);
+    }
+
+    /// @custom:deprecated Use terminateService(uint256,bytes) instead
+    function terminateService(uint256 dataSetId) public {
+        _terminateService(dataSetId, "");
+    }
+
+    function _terminateService(uint256 dataSetId, bytes memory extraData) private {
```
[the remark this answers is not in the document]
wjmelements: That duplicates the `_terminateService` logic.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -98,7 +104,7 @@ contract FilecoinWarmStorageService is
     );
 
     event ServiceTerminated(
```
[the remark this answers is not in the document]
wjmelements: I actually instructed claude to add an event and they just changed the prior one in b0640771e18e4597604d95edab9933d9340e2620. I didn't notice that the event already existed. By the time I made the change you are reviewing I thought I was modifying that new event.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: I don't know what `transaction` means in the context of subgraph. Does it capture internal transactions? That is, is it `origin()` or `caller()`?

I don't think we should use `address(0)`. The prior behavior is to use `msg.sender` in the non-signature case, and we should keep that if we can, for backwards compatibility.

The difficulty of distinguishing consent is that the client can sign either the message or be the sender. So we need that `from` field to distinguish these cases. Then consensual termination is `ServiceTerminated.signer != SP && transaction.from == SP`.

service_contracts/src/FilecoinWarmStorageService.sol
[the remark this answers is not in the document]
wjmelements: Going to revert the signature change.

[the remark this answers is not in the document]
wjmelements: 27dc618

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: 27dc618

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -98,7 +104,11 @@ contract FilecoinWarmStorageService is
...
```
rvagg, line 113:
```suggestion
    /// @notice Emitted when a data set's service is terminated.
    /// @dev `approver` authorizes the termination: the recovered signer when extraData
    ///      carries a signature (the payer, or a session key resolving to the payer via
    ///      SessionKeyRegistry), otherwise the direct caller.
    //       Combine with tx.origin (the account that sent the transaction) to classify:
    ///        - approver == serviceProvider ->  SP, unilateral termination
    ///        - else                        ->  client-authorized termination; mutual if
    ///                                          tx.origin == serviceProvider, else client unilateral
    ///                                          (approver != payer means a session key signed.)
    event ServiceTerminated(
```

Now that `approver` is used to derive one of 3 conditions we really need to be clear about it for anyone consuming it, it's not obvious at all what the signalling is here.

rvagg (approved):
Approving on the condition that we get docs in about how to properly make use of `ServiceTerminated` now it's a reading-tea-leaves exercise. There's 3 distinct termination paths, and there's the session key layered on top of two of those paths.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: `event.transaction.from` is `tx.origin` (`ORIGIN`) but the relevant information is `msg.sender` which is sometimes different, but (I believe) unavailable in subgraph.

wjmelements: 0d155b0d8ae49496673b7d72bfb19781e0b945ae

PR #495, opened 2026-06-01
fix: trigger Synapse SDK update on published releases

rjan90:
## Summary

- Trigger the Synapse SDK update workflow when a GitHub Release is published.
- Keep the existing tag-push and manual dispatch paths.
- Resolve the release tag from the release event payload when applicable.

## Why

The `v1.2.1` release publication emitted a `release: published` event, but the workflow only listened for tag pushes and manual dispatches, so no Synapse SDK PR was created.

### DIFF
--- a/.github/workflows/notify-synapse-sdk.yml
+++ b/.github/workflows/notify-synapse-sdk.yml
@@ -1,6 +1,8 @@
 name: Update Synapse SDK
 
 on:
+  release:
+    types: [published]
   push:
     tags:
       - 'v*'
@@ -55,12 +57,17 @@ jobs:
           EVENT_NAME: ${{ github.event_name }}
           REF_NAME: ${{ github.ref_name }}
           REF_TYPE: ${{ github.ref_type }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
           INPUT_TAG: ${{ inputs.tag }}
           INPUT_SHA: ${{ inputs.sha }}
         run: |
           set -euo pipefail
 
           tag="$REF_NAME"
+          if [ "$EVENT_NAME" = "release" ] && [ -n "$RELEASE_TAG" ]; then
+            tag="$RELEASE_TAG"
+          fi
+
           if [ "$EVENT_NAME" = "workflow_dispatch" ] && [ -n "$INPUT_TAG" ]; then
             tag="$INPUT_TAG"
           fi


rjan90:
Manual backfill dispatch for `v1.2.1` was run after opening this PR: https://github.com/FilOzone/filecoin-services/actions/runs/26751117563.

Outcome: the workflow started correctly but failed before pushing/opening the Synapse SDK PR during `pnpm install` in `FilOzone/synapse-sdk` : `ERR_PNPM_TRUST_DOWNGRADE` for `tinyexec@1.2.2`, pulled while installing `astro@6.3.7` under `docs`. So no Synapse PR was created yet. The release-trigger fix in this PR still addresses the original missed trigger; the current backfill blocker is the Synapse SDK dependency trust policy and should be resolved there, then this dispatch can be rerun.

rjan90:
Follow-up on the failed manual backfill dispatch, for Hugo to review.

The filecoin-services workflow itself now starts correctly, but the run fails inside `FilOzone/synapse-sdk` during dependency install. I reproduced the next steps in a temp `synapse-sdk` clone.

Current blockers:

1. `trustPolicy: no-downgrade` rejects `tinyexec@1.2.2` with `ERR_PNPM_TRUST_DOWNGRADE`.
2. After excluding that package, `blockExoticSubdeps` rejects `viem@2.51.0` because that release depends on `ox` through `https://pkg.pr.new/...`.

The suggested Synapse SDK-side patch is to keep the supply-chain guards enabled, but add a narrow trust-policy exclusion for the reviewed `tinyexec` versions and pin `viem` below the release with the exotic transitive dependency:

```yaml
catalog:
  viem: 2.50.4

trustPolicyExclude:
  - tinyexec@1.2.2 || 1.2.3 || 1.2.4
```

Important: `viem` should be pinned exactly as `2.50.4`, not `^2.50.4`, otherwise pnpm can float back to `2.51.0+` and hit the exotic `ox` URL again.

Validation from the temp clone with the above changes plus the `v1.2.1` filecoin-services ref:

```bash
pnpm install
pnpm -r --filter @filoz/synapse-core run generate-abi
pnpm run lint
pnpm run build
```

All four passed locally. Once that Synapse SDK dependency-policy fix lands, we should rerun the `v1.2.1` manual dispatch from this repo.

BigLep:
The workflow trigger fix here looks good, but the manual backfill dispatch is blocked until the synapse-sdk dependency policy issues are resolved. Tracked in FilOzone/synapse-sdk#809 (tinyexec trust-policy exclusion + viem pin). Once that lands, the v1.2.1 dispatch can be rerun.

rjan90:
Update: reran the manual `v1.2.1` backfill after `FilOzone/synapse-sdk#810` landed.

Result: successful.

- Workflow run: https://github.com/FilOzone/filecoin-services/actions/runs/26805761691
- Resolved tag/SHA: `v1.2.1` / `4e548903095cfb46bc35af029f2ae0f39f18b8e4`
- Synapse branch: `chore/update-filecoin-services-v1.2.1`
- Opened Synapse SDK PR: https://github.com/FilOzone/synapse-sdk/pull/812

The previous Synapse SDK dependency-policy blocker is resolved. The only annotation on the successful run is the existing GitHub Actions Node 20 deprecation warning.

hugomrdias (approved):

PR #497, opened 2026-06-02
feat(pricing): set createDataSet fee to $0.025

rvagg:
With the sybil burn removed, this fee is now the economic friction against frivolous dataset creation. 10x keeps it negligible for legitimate use (single-copy $0.025, 2-copy $0.05) while meaningfully compensating the SP and adding a slight behavioural disincentive.


wjmelements (approved):

rjan90 (approved):

ZenGround0:
The return of the sybil fee !
👹

PR #498, opened 2026-06-02
refactor(pricing): rename "minimum" to "required" in payer validation

rvagg:
Cosmetic, for clarification purposes since I noticed that some of the old floor price language carried over so I want to clear that up. Shouldn't change the event signatures even though the param names are changing.

Pre-flight checks aren't enforcing minimum charges, they verify the payer has approved capacity to add pieces later. Empty datasets leave the per-dataset fee headroom untouched; required up front intentionally so creation assumes pieces can follow. Error selectors unchanged (computed from types, not names).

### DIFF
--- a/service_contracts/abi/Errors.abi.json
+++ b/service_contracts/abi/Errors.abi.json
@@ -286,7 +286,7 @@
         "internalType": "uint256"
       },
       {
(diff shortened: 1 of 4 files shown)


rjan90 (approved):
LGTM!

PR #499, opened 2026-06-02
fix(metadata): remove addPieces extraData byte cap, adjust limits

rvagg:
Drop the FWSS addPieces extraData size limit since batch size is bounded upstream by PDPVerifier/FVM event size.

Tighten piece metadata limits to 3 keys per piece and 96 bytes per value, and update tests for the new boundaries. This keeps gas low enough to provide a (relatively) stable 40 piece limit even with metadata; limit can be hard-applied in SP & SDK software for clearer errors to users.

Ref: https://app.notion.com/p/filecoindev/addPieces-Batch-Limits-and-Proposed-Changes-36ddc41950c180d39f45f0da8982090c

### DIFF
--- a/service_contracts/test/FilecoinWarmStorageService.t.sol
+++ b/service_contracts/test/FilecoinWarmStorageService.t.sol
@@ -11,7 +11,6 @@ import {SessionKeyRegistry} from "@session-key-registry/SessionKeyRegistry.sol";
 
 import {
     CHALLENGES_PER_PROOF,
-    MAX_ADD_PIECES_EXTRA_DATA_SIZE,
     MAX_CREATE_DATA_SET_EXTRA_DATA_SIZE,
     FilecoinWarmStorageService
 } from "../src/FilecoinWarmStorageService.sol";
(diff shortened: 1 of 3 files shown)


rvagg:
Note that the lingering concern about create+add gas costs pushing us beyond the 40 piece limit may not be a binding concern now with the sybil rail removal from createDataSet. We'll have to measure that one again to see if 40 is the right number or should be lower.

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -11,8 +11,6 @@ import {SessionKeyRegistry} from "@session-key-registry/SessionKeyRegistry.sol";
 
 import {
```
rjan90: Looks like CI wants this collapsed into one line

rjan90 (approved):
Only the lint/formatting issue needs to be addressed: the multiline import was collapsed onto one line. Otherwise, the change looks good to me.

hugomrdias (approved):

PR #500, opened 2026-06-02
feat: cleanup abandoned data set

### LINKED ISSUE #484: FWSS support for simplified DataSet terminations and cleanup for GA
https://www.notion.so/filecoindev/Data-Set-Terminations-Clean-up-360dc41950c1801ebf0aff017a322e7f

## Summary

Implementation of the FWSS (FilecoinWarmStorageService) contract changes for the data set termination and clean-up design. This is the filecoin-services side of the work; see sibling issues for other repos.

### 712 typehash rename

- [x] `DeleteDataSet(uint256 dataSetId)` → `TerminateService(uint256 dataSetId)` in test fixtures and signature verification helpers (#480).
- [x] Move `TERMINATE_SERVICE_TYPEHASH` from the test contract into `SignatureVerificationLib` (production code) when the `terminateService` impl lands. (https://github.com/FilOzone/filecoin-services/pull/485)

### `terminateService` changes

- [x] Accept `extraData` carrying client EIP-712 signature for SP-relayed termination. (https://github.com/FilOzone/filecoin-services/pull/485)
- [ ] When valid client 712 present: collect client termination fee via one-time payment on existing rail, call `modifyRailLockup(railId, 0, 0)` to drop lockup period to zero for immediate finalisation. (separate issue TBD)
- [x] When `extraData` empty: SP-initiated termination, 30-day lockup path, no client fee. (https://github.com/FilOzone/filecoin-services/pull/488)
- [x] Move CDN rail termination out of `terminateService` — CDN rails now persist through the lockup window and terminate in `dataSetDeleted` instead. (https://github.com/FilOzone/filecoin-services/pull/488)

### `dataSetDeleted` callback — two-path listener

- [x] **Normal path** (SP calls, rail already terminated and finalised): existing cleanup, plus CDN rail termination moves here from `terminateService`. (https://github.com/FilOzone/filecoin-services/pull/488)
- [ ] **Abandonment path** (`pdpEndEpoch == 0`, 30d since last proof, any caller): FWSS performs inline rail tear-down (`modifyRailLockup` → `terminateRail` → `settleRail`), handles both funded and underfunded client states, then clears FWSS state. (separate issue TBD)

### CDN rail lifecycle change

- [x] CDN service persists through the storage lockup window (client keeps retrieval acceleration for what they paid for). (#488)
- [x] CDN rails terminated in `dataSetDeleted` with a shorter lockup period (5 days) to give FilBeam a graceful settle window. (#488)
- [x] Existing FilBeam rails with 30-day lockup left as-is; new ones created with shorter period. (#488)

### Client termination fee

- [x] Collected at `terminateService` when client 712 is present (client is provably involved). (#488)
- [x] Fee compensates SP for gas; no link to state clean-up incentives. (#488)

## Sibling issues

- PDPVerifier (cleanup bond, `cleanupPieces`): https://github.com/FilOzone/pdp/issues/269
- Curio (new tasks, HTTP endpoint, reconciliation): https://github.com/filecoin-project/curio/issues/1231
- Synapse SDK (712 flow, pre-check, typehash rename): https://github.com/FilOzone/synapse-sdk/issues/792
- Design issue: https://github.com/FilOzone/filecoin-services/issues/477

wjmelements:
Reviewers @zenground0 @rvagg
Toward #484
This implements the abandonment path, which allows deleting the data set without `terminateService` if the rail has not been proven recently.
The motivation for this is that we want to allow cleanup.
Part of the abandonment logic (the gating and the deposit) is already implemented in `PDPVerifier`.
We need these additional `_verifyInactivity` checks to prevent the SP from skipping `terminateService` in the non-abandonment case.
#### Changes
* `_verifyInactivity`
* `Rails.abandonRails`
* tests

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -34,6 +34,8 @@ import {SignatureVerificationLib} from "./lib/SignatureVerificationLib.sol";
 uint256 constant NO_PROVING_DEADLINE = 0;
 uint64 constant CHALLENGES_PER_PROOF = 5;
 uint256 constant COMMISSION_MAX_BPS = 10000; // 100% in basis points
+// Mirror of PDPVerifier.INACTIVITY_WINDOW (lib/pdp/src/PDPVerifier.sol)
+uint256 constant PDP_INACTIVITY_WINDOW = 86400;
 
 /*
 * Maximum extraData for createDataSet
@@ -117,6 +119,8 @@ contract FilecoinWarmStorageService is
         uint256 cacheMissRailId,
         uint256 cdnRailId
     );
+
+    event DataSetAbandoned(uint256 indexed dataSetId, uint256 pdpRailId, uint256 cacheMissRailId, uint256 cdnRailId);
 
     event PDPPaymentTerminated(uint256 indexed dataSetId, uint256 endEpoch, uint256 pdpRailId);
 
@@ -621,8 +625,10 @@ contract FilecoinWarmStorageService is
     }
 
     /**
-     * @notice Handles data set deletion after the payment rails were terminated
-     * @dev Called by the PDPVerifier contract when a data set is deleted
+     * @notice Handles data set deletion after voluntary termination or via the abandonment path.
+     * @dev Called by the PDPVerifier contract when a data set is deleted.
+     *      When pdpEndEpoch == 0 the rail was never terminated via terminateService; FWSS verifies
+     *      30-day inactivity and performs inline teardown so a keeper needs only one transaction.
      * @param dataSetId The ID of the data set being deleted
      */
     function dataSetDeleted(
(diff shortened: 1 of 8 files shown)


service_contracts/test/Abandonment.t.sol
```diff
@@ -0,0 +1,305 @@
...
+        vm.deal(sp, 10 ether);
+        vm.deal(client, 10 ether);
+        vm.deal(keeper, 10 ether);
+
+        usdfc = new MockERC20();
+        payments = new FilecoinPayV1();
+        sessionKeyRegistry = new SessionKeyRegistry();
+
+        // Deploy real PDPVerifier as a proxy
+        PDPVerifier pdpVerifierImpl = new PDPVerifier(1, 0);
```
wjmelements, line 58: other tests do not use the real PDPVerifier, but I think we should here

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1445,12 +1401,13 @@ contract FilecoinWarmStorageService is
...
 
-        // If proving wasn't ever activated for this data set, don't pay anything
+        // If proving wasn't ever activated for this data set, don't pay anything.
+        // Return settleUpto = toEpoch (not fromEpoch) so settlement can still advance and discharge lockup.
         uint256 activationEpoch = provingActivationEpoch[dataSetId];
         if (activationEpoch == 0) {
             return ValidationResult({
                 modifiedAmount: 0,
-                settleUpto: fromEpoch,
+                settleUpto: toEpoch,
```
wjmelements, line 1416: This change is necessary to allow the rail to settle; otherwise it can only be finalized without validation, which would pay the SP for unproven periods. `settleRail` would revert with `NoProgressInSettlement`.

wjmelements: I'm going to try to determine the intent of the prior logic

wjmelements: I haven't found any good reason for the prior behavior.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -703,6 +691,22 @@ contract FilecoinWarmStorageService is
         delete dataSetInfo[dataSetId];
     }
 
+    /**
+     * @notice Verifies that a data set has been inactive for the required inactivity window.
+     * @dev Reverts if the SP has been active within PDPVerifier.INACTIVITY_WINDOW epochs.
+     *      Never-activated datasets (provingActivationEpoch == 0) are unconditionally accepted.
```
rvagg:
```suggestion
     *      Never-activated datasets (provingActivationEpoch == 0) are unconditionally accepted,
     *      as PDPVerifier.deleteDataSet already performs its own since-activiation check.
```

service_contracts/src/FilecoinWarmStorageService.sol
rvagg: seems reasonable to me, I can't think through a case with `activationEpoch == 0` where this might be a problem

rvagg (approved):
@wjmelements I've pushed my own docs commit in here, I had to work through some of the changes and figured that docs to clear up ambiguity would be helpful to future readers. Adjust if you disagree with anything I've said in there.

rvagg:
and another commit for SPEC.md

PR #506, opened 2026-06-03
feat(terminateService): immediate termination in the consent case

### LINKED ISSUE #484: FWSS support for simplified DataSet terminations and cleanup for GA

wjmelements:
Reviewers @zenground0 @rvagg
If both the SP and the client agree, termination is immediate.
The final lockup period is 0, and deletion and cleanup can proceed.
Perhaps we should also `settleRail` in this case, but I don't currently think that's beneficial because of the case where the rail is far behind.
I pass the flag down to the Rails library to ensure that we only `modifyRailLockup` once.
#### Changes
* implement immediate termination
* update doc and tests

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -1006,6 +1006,7 @@ contract FilecoinWarmStorageService is
         require(info.pdpEndEpoch == 0, Errors.DataSetPaymentAlreadyTerminated(dataSetId));
 
         address approver;
+        bool immediate = false;
         if (extraData.length > 0) {
             require(
                 extraData.length <= MAX_TERMINATE_SERVICE_EXTRA_DATA_SIZE,
@@ -1014,7 +1015,7 @@ contract FilecoinWarmStorageService is
             bytes memory signature = abi.decode(extraData, (bytes));
             approver = _verifyTerminateServiceSignature(info.payer, dataSetId, signature);
             if (msg.sender == info.serviceProvider) {
-                // TODO termination can be immediate
+                immediate = true;
                 info.pendingOneTimePayments += uint96(TERMINATE_FEE);
             }
         } else {
@@ -1033,6 +1034,9 @@ contract FilecoinWarmStorageService is
             updatePaymentRates(dataSetId, info, leafCount, pending, info.lifecycleReserveBalance);
         }
 
+        if (immediate) {
+            payments.modifyRailLockup(info.pdpRailId, 0, 0);
+        }
         payments.terminateRail(info.pdpRailId);
 
         emit ServiceTerminated(approver, dataSetId, info.pdpRailId, info.cacheMissRailId, info.cdnRailId);
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -655,7 +655,7 @@ contract FilecoinWarmStorageService is
         } else {
             // Normal path: terminateService was already called.
             // Verify the payment window has elapsed and the rail is fully settled.
-            require(block.number > info.pdpEndEpoch, Errors.PaymentRailsNotFinalized(dataSetId, info.pdpEndEpoch));
+            require(block.number >= info.pdpEndEpoch, Errors.PaymentRailsNotFinalized(dataSetId, info.pdpEndEpoch));
             try payments.getRail(info.pdpRailId) returns (FilecoinPayV1.RailView memory rail) {
                 require(
                     rail.settledUpTo >= rail.endEpoch,
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -841,7 +841,7 @@ contract FilecoinWarmStorageService is
 
         uint96 newPending = info.pendingOneTimePayments + uint96(SCHEDULE_PIECE_REMOVALS_FEE);
         info.lifecycleReserveBalance = FilecoinPayV1(paymentsContractAddress).replenishReserveIfNeeded(
-            info.pdpRailId, info.pdpEndEpoch, info.lifecycleReserveBalance, newPending, false
+            info.pdpRailId, info.pdpEndEpoch, info.lifecycleReserveBalance, newPending
         );
         info.pendingOneTimePayments = newPending;
 
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -774,7 +774,7 @@ contract FilecoinWarmStorageService is
 
         // Validate lockup for the new data set size (fail-fast if client has insufficient funds)
         uint256 currentLeafCount = IPDPVerifier(pdpVerifierAddress).getDataSetLeafCount(dataSetId);
-        updatePaymentRates(dataSetId, info, currentLeafCount, pending, reserveBalance);
+        updatePaymentRates(dataSetId, info, currentLeafCount, pending, reserveBalance, false);
 
         // Store metadata for each new piece
         for (uint256 i = 0; i < pieceData.length; i++) {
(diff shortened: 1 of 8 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -655,7 +655,7 @@ contract FilecoinWarmStorageService is
         } else {
             // Normal path: terminateService was already called.
             // Verify the payment window has elapsed and the rail is fully settled.
-            require(block.number > info.pdpEndEpoch, Errors.PaymentRailsNotFinalized(dataSetId, info.pdpEndEpoch));
+            require(block.number >= info.pdpEndEpoch, Errors.PaymentRailsNotFinalized(dataSetId, info.pdpEndEpoch));
```
wjmelements, line 658: this allows batching

service_contracts/src/lib/Rails.sol
rvagg, line 258: This is misnamed now there's this hidden feature that does something entirely different, `prepareReserveForFlush` might work better?
Or, since we only call this in one place, how about just a simple function to do that 1 line and then pull the `if` into `updateStorageRates`.

rvagg: actually since `updateStorageRates` is in here, just `if` it in there and avoid the extra arg and complexity in here

rvagg: because the other place it's called, this arg is always `false`; so I think the case for only doing  the special dance in `updateStorageRates` is pretty strong

service_contracts/test/OpFees.t.sol
```diff
@@ -202,15 +202,24 @@ contract OpFeesTest is FilecoinWarmStorageServiceTest {
         uint96 reserveBefore = info.lifecycleReserveBalance;
         assertEq(info.pendingOneTimePayments, 0);
 
+        (, uint256 spFundsBefore,,) = payments.getAccountInfoIfSettled(mockUSDFC, sp1);
+
```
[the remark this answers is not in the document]
rvagg: 👍 
what's the good of linting if it can't pick this kind of thing up?

service_contracts/src/lib/Rails.sol
```diff
@@ -260,8 +260,13 @@ library Rails {
         uint256 pdpRailId,
         uint256 pdpEndEpoch,
         uint96 reserveBalance,
-        uint96 pending
+        uint96 pending,
+        bool immediateTermination
     ) internal returns (uint96) {
+        if (immediateTermination) {
+            payments.modifyRailLockup(pdpRailId, 0, pending);
```
rvagg: awkward edge case of a client doing some pre-termination ops to raise the `pending` above replenish threshold, draining their reserve and also their FP balance, requesting immediate termination and this reverting .. do we care? can we even do anything to provide an informative error in the revert? or just live with it?

probably live with it.. could document it, seems very edge-casey although someone going through and performing bulk cleanup to get themselves out of FP/FOC entirely might find themselves locked from doing this operation.

rvagg (approved):
very nice, just the `replenishReserveIfNeeded` awkwardness I wouldn't mind addressed, I think just move the `immediateTermination` case up in to `updateStorageRates`, swallow the slight increase in LOC there and `replenishReserveIfNeeded` remains focused and well-named

service_contracts/test/OpFees.t.sol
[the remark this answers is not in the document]
wjmelements: the compiler caught it though

409f000d3552731d34924b88907acd573a591d3c

service_contracts/test/OpFees.t.sol
```diff
@@ -331,17 +340,25 @@ contract OpFeesTest is FilecoinWarmStorageServiceTest {
         FilecoinWarmStorageService.DataSetInfoView memory info = viewContract.getDataSet(dataSetId);
         uint96 reserveAfterAdd = info.lifecycleReserveBalance;
 
+        (, uint256 spFundsBefore,,) = payments.getAccountInfoIfSettled(mockUSDFC, sp1);
+
```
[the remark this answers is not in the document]
wjmelements: 409f000d3552731d34924b88907acd573a591d3c

service_contracts/src/lib/Rails.sol
wjmelements: 2bebfa1573a395403b0785c26b0d00c8414ce86e

service_contracts/src/lib/Rails.sol
wjmelements: I don't think so; I think this can only reduce the lockup

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1014,7 +1015,7 @@ contract FilecoinWarmStorageService is
             bytes memory signature = abi.decode(extraData, (bytes));
             approver = _verifyTerminateServiceSignature(info.payer, dataSetId, signature);
             if (msg.sender == info.serviceProvider) {
```
wjmelements: I think we should require `msg.sender == info.serviceProvider` to prevent frontrunning

service_contracts/src/lib/Rails.sol
wjmelements: ah this is https://github.com/FilOzone/filecoin-pay/issues/290

wjmelements: Nothing we can do about it, unfortunately.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: 9c09f786c19fd9abf779229774e78c0262874580

rvagg (approved):
new changes are fine, it does close the opportunity for the session-key delegated party doing terminations separate to the SP (e.g. an MSP style delegate), so an SP can hold data hostage, but maybe we just deal with that when we get there cause we can always go backward on this one

service_contracts/src/lib/Rails.sol
wjmelements: I did a workaround in #520

PR #520, opened 2026-06-10
fix: allow abandonment and consensual termination of underfunded data sets

### LINKED ISSUE #519: Support underfunded termination
A data set is underfunded if its payer cannot afford to `settleAccountLockup`. Such data sets cannot even decrease their lockup period or their payment rate, as documented in https://github.com/FilOzone/filecoin-pay/issues/290. Therefore, underfunded termination fails with `LockupPeriodChangeNotAllowedDueToInsufficientFunds` in the two cases where we try to set endEpoch to the current block:
1. abandonment path in `PDPVerifier.deleteDataSet`
2. consensual termination path in `FilecoinWarmStorageService.terminateService`

wjmelements:
Reviewer @rvagg
Fixes #519
Consensual termination and abandonment both attempt to set the endEpoch to block.number in order to allow immediate cleanup.
Because we cannot modify the payment rate or the lockup for underfunded data sets before termination (see https://github.com/FilOzone/filecoin-pay/issues/290), we are unable to reduce the lockup period to ensure endEpoch is the current block.
Instead of reverting, we will release the fixed lockup after termination.
#### Changes
* add three failing tests
* catch modifyRailLockup exceptions in abandonment and consensual termination


service_contracts/src/lib/Rails.sol
```diff
@@ -286,7 +295,13 @@ library Rails {
...
         uint256 newStorageRatePerEpoch = calculateStorageRate(leafCount);
         if (immediateTermination) {
-            payments.modifyRailLockup(pdpRailId, 0, pending);
+            // Try to zero the lockup period (requires a fully-settled account).
+            // For underfunded payers the period change is blocked.
+            try payments.modifyRailLockup(pdpRailId, 0, pending) {}
+            catch {
+                // Always release the fixed lockup
+                payments.modifyRailLockup(pdpRailId, DEFAULT_LOCKUP_PERIOD, pending);
+            }
```
rvagg: I don't think we should do this here, immedaiteTermination path implies that everyone is behaving properly and agreeing together, allowing an underfunded client here to proceed with immediate termination gives a bad actor privileges that I don't think they should get.

In the normal case I don't think we'll get here, as I said in #519 I think the only way we can get into the situation where this reverts is where the SP abandons the data set and doesn't discover underfunding with regular settles. The immediateTermination path suggests the SP is still engaged and willing to submit a msg to the chain for this data set.

I'd just leave a comment in here about why we're not doing a `try/catch`.

```suggestion
            // No try/catch here to cover the underfunded-client case. immediateTermination implies
            // the payer is solvent enough to consent to releasing the lifecycle reserve and paying the
            // terminate fee from `pending` in one step.
            payments.modifyRailLockup(pdpRailId, 0, pending);
```

wjmelements: This makes sense to me. I'll think about it a bit more.

wjmelements: b7b39ac315ae8ec23960cba88731e5f417a8bab8

service_contracts/src/lib/Rails.sol
```diff
@@ -182,18 +180,29 @@ library Rails {
...
         payments.settleRail(pdpRailId, block.number);
-        payments.modifyRailLockup(pdpRailId, 0, 0);
+
+        // Try to zero the lockup period before termination.
+        // For underfunded payers the period change will fail.
+        bool reserveRemaining = false;
+        try payments.modifyRailLockup(pdpRailId, 0, 0) {}
+        catch {
+            reserveRemaining = true;
+        }
```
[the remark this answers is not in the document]
wjmelements: Dug into this — at the actual call site (`payments.modifyRailLockup(pdpRailId, 0, 0)` in `abandonRails`), I don't think any other revert reason is reachable through the real flow:

- `dataSetDeleted` only routes into `abandonRails` when `info.pdpEndEpoch == 0`, so the PDP rail entering this call is always fresh/non-terminated — `modifyTerminatedRailLockup`'s branch (and its `InvalidTerminatedRailModification` error) can't fire here.
- The PDP rail's validator (`railTerminated`) requires `terminator == address(this)`, so nothing but FWSS itself can ever terminate it ahead of time — unlike CDN rails, the payer/operator can't pre-terminate it externally.
- With `period=0, lockupFixed=0` on a non-terminated rail, `modifyNonTerminatedRailLockup` either passes (fully-settled payer) or reverts on the very first check with `LockupPeriodChangeNotAllowedDueToInsufficientFunds` (not fully settled) — the fixed-lockup-increase check, operator-max-period check, and `CurrentLockupLessThanOldLockup` invariant can't trigger for a same-or-decreasing `(0, 0)` call.

So narrowing the catch to only swallow `LockupPeriodChangeNotAllowedDueToInsufficientFunds` would be a no-op in practice today, and I couldn't construct a real test that hits a different revert without manually corrupting storage. Happy to add the narrowing defensively anyway if you'd like, but I don't have evidence it currently masks anything.

— posted by Claude

service_contracts/src/lib/Rails.sol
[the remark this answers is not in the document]
wjmelements: this path no longer does try-catch

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ See the [v1.3.0 GitHub Release](https://github.com/FilOzone/filecoin-services/re
 
 ### Fixed
 - Fixed abandonment cleanup for activated data sets so `PDPVerifier.deleteDataSet` can complete final settlement after terminating rails, and cleared proving-period storage during data-set deletion ([#512](https://github.com/FilOzone/filecoin-services/pull/512)).
+- Fixed `abandonRails` to handle underfunded payers: when zeroing the lockup period is blocked by FilecoinPay, the lifecycle reserve is still released immediately but the PDP rail retains its default 30-day lockup period rather than finalizing at once ([#520](https://github.com/FilOzone/filecoin-services/pull/520)).
```
rvagg: this is under 1.3.0, we need an "Unreleased" section

service_contracts/test/Abandonment.t.sol
```diff
@@ -419,6 +425,113 @@ contract AbandonmentTest is MockFVMTest {
...
+
+        _topUpAndSettle(payer, pdpRailId, 1);
+
+        vm.roll(vm.getBlockNumber() + (PDP_INACTIVITY_WINDOW - 1));
+
+        vm.prank(keeper);
+        pdpVerifier.deleteDataSet(dataSetId, "");
+
+        _assertDataSetCleared(dataSetId);
+        assertEq(payments.getRail(pdpRailId).lockupFixed, 0, "lifecycle reserve should be zero after abandonment");
```
rvagg: @wjmelements I think we might have a stranded rail situation here; the rail is terminated (no more lockup) but its unsettled _span_ keeps `paymentRate * (endEpoch-settledUpTo)` locked in the payers account. Because of `deleteDataSet` this rail isn't FWSS associated anymore so can't be settled. So a client ends up with stranded funds that can't be settled. They could call `settleTerminatedRailWithoutValidation` but that settles in the SP's favour which doesn't help unblock funds that should go back to the payer.

```solidity
// tail is stranded in account lockup: (not in available funds)
(, uint256 funds, uint256 available,) = payments.getAccountInfoIfSettled(usdfc, payer);
uint256 tail = payments.getRail(pdpRailId).paymentRate * (payments.getRail(pdpRailId).endEpoch - block.number);
assertEq(available, funds - tail);   // tail locked

// bricked
vm.roll(payments.getRail(pdpRailId).endEpoch + 1);
vm.expectRevert(abi.encodeWithSelector(Errors.RailNotAssociated.selector, pdpRailId));
payments.settleRail(pdpRailId, block.number);   // validator callback reverts
```

From what I can see there might be a couple of ways of dealing with this:

1. Retain the `railToDataSet`, even though we don't have a good trigger to removing it eventually, perpetually dead state, maybe a cost we should buy. Could do it after abandonment, if you `payments.getRail` on the rail and it reverts then it's good to clean up, but if it doesn't revert then there's still funds in there, we can retain it.
2. Be lax in `validatePayment` and accept that if we don't have a `railToDataSet` entry for this `railId` then it was abandoned and left dangling, in which case we just settle it in the client's favour automatically, no questions. I _think_ this is safe, it's the only case we'd be called for a data set we don't own.

wjmelements: Good catch. We don't want these funds to go to the SP in the abandonment case. I am implementing the second idea. It's somewhat difficult to describe in the docstring so I am iterating on it with Claude.

service_contracts/src/lib/Rails.sol
```diff
@@ -167,11 +167,12 @@ library Rails {
...
-    /// @dev SP forfeits pending op-fees; lifecycle reserve and streaming buffer return to the
-    ///      payer. Intentional: abandonment means the SP walked away.
-    ///      PDP rail flow: settle to pay any proven epochs and advance settledUpTo; zero the
-    ///      lockup to release buffer+reserve to the payer; terminate (endEpoch = block.number
-    ///      since period is 0); settle again so settledUpTo >= endEpoch finalises the rail.
+    /// @dev SP forfeits pending op-fees; lifecycle reserve returns to the payer.
+    ///      For well-funded payers the rail is terminated with no lockup period, releasing the
+    ///      reserve immediately.
+    ///      For underfunded payers the PDP rail remains for DEFAULT_LOCKUP_PERIOD; the payer's
+    ///      reserve isn't released until that period elapses, though every proven epoch is paid
```
rvagg:
```suggestion
    ///      streaming lockup tail isn't released until that period elapses, though every proven epoch is paid
```

"reserve" (per the way we're using it in here) is released immediately, it's the streaming lockup that's not

SPEC.md
```diff
@@ -240,13 +240,13 @@ require(settledUpTo >= endEpoch, RailNotFullySettled)
...
+**Abandonment path**: When `pdpEndEpoch == 0` (the SP never called `terminateService`), the data set can still be deleted once inactive for `INACTIVITY_WINDOW` (30 days from `lastProvenEpoch`, or from `provingActivationEpoch` for activated-but-never-proven data sets). PDPVerifier gates this: SP-only within the window, permissionless after. FWSS layers its own `_verifyInactivity` check on top so the SP cannot use this path to skip `terminateService` on an active data set. Inline teardown via `Rails.abandonRails` settles the PDP rail (advancing through unproven epochs via the pre-activation short-circuit), releases the lifecycle reserve back to the payer, terminates the rail, and best-efforts the CDN rails. The SP forfeits any pending one-time op-fees; this is intentional, since the SP walked away. For well-funded payers the lockup period is zeroed before termination, releasing the streaming buffer immediately. For underfunded payers the lockup period cannot be zeroed, so the PDP rail retains its default 30-day window and the streaming buffer is released only after that window elapses.
```
rvagg, line 249: could add an extension to this para about how it works now

rvagg (approved):
nice; couple of doc nits, but that CHANGELOG correction needs to be taken care of - don't edit the 1.3.0 section, make a new Unreleased section

rjan90:
Is there anything blocking merging this?

BigLep:
@rjan90 : I think we incorporate Rod's doc nits and then merge.

service_contracts/test/Abandonment.t.sol
wjmelements: [587dab4](https://github.com/FilOzone/filecoin-services/pull/520/commits/587dab46be983fcd6628cc7cf8a1a3c073afb5a2)

PR #521, opened 2026-06-11
fix: prevent lifecycle reserve underflow

wjmelements:
Reviewer @zenground0
This fixes a bug with the lifecycle reserve where subtraction underflow can cause nextProvingPeriod to fail.
#### Changes
* add failing test
* prevent underflow in `updateStorageRates`


ZenGround0 (approved):
Very nice find

PR #522, opened 2026-06-11
fix: publish FWSS library ABIs

rvagg:
Rails library in particular has event signatures, and I've just discovered that we don't have these published ABI artifacts. I guess we've missed the 1.3.0 boat but we should remember for next time - any non-empty library we extract should have ABI artifcats included.

### DIFF
--- a/service_contracts/Makefile
+++ b/service_contracts/Makefile
@@ -135,7 +135,7 @@ check-layout:
 
 # ABI Management
 
-# Core contracts we publish ABIs for
+# Core contracts and externally consumed libraries we publish ABIs for.
 ABI_CONTRACTS := \
 	FilecoinWarmStorageService \
 	FilecoinWarmStorageServiceStateLibrary \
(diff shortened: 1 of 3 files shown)


wjmelements:
This creates a separate abi.json for the libraries but ideally we would want the events from those libraries included in the ABI of the implementation that links them. This is because web3 libraries reading ABI will assume the events were emitted from the contract address with the ABI, but libraries are accessed via `DELEGATECALL`. So we want to merge library events into the abi.json of the contracts that use them, and discard other library fields such as their method ABI, which are useless externally.

We could keep them separate like this but then the merging would need to happen in all of the consumers. I think it best we merge the ABIs in one place.

A similar problem occurs with proxy ABI; what these have in common is `DELEGATECALL`.

BigLep:
@rvagg : what are the next steps here? Are you planning to take them or should we ask someone else?

rvagg:
Rebased and updated to merge into one file. I don't love the bash & jq magic in here but it is what it is. PTAL @wjmelements

service_contracts/Makefile
```diff
@@ -148,33 +148,37 @@ ABI_CONTRACTS := \
...
 # Directory for ABIs
 abi:
 	@mkdir -p abi
 
 # Update ABIs
 .PHONY: update-abi
-update-abi: $(ABI_FILES)
+update-abi: | abi
+	forge clean
+	forge build --via-ir
```
wjmelements: Do we have to do a clean build? This takes forever!

service_contracts/Makefile
```diff
@@ -148,33 +148,37 @@ ABI_CONTRACTS := \
...
+	@rm -f abi/*.abi.json
+	@set -e; for contract in $(ABI_CONTRACTS); do \
+		artifact="out/$$contract.sol/$$contract.json"; \
+		echo "Extracting ABI for $$contract..."; \
+		jq '.abi' "$$artifact" > "abi/$$contract.abi.json"; \
+	done
```
wjmelements, line 169: This replaces the ABI_RULE but it seems the new logic is only used for one contract's ABI. Perhaps we can use make's `filter-out` for the special case and use the combination logic for only the FWSS contract.

service_contracts/Makefile
wjmelements: For `gen` we have `clean-gen`, so perhaps we could follow that pattern.

rvagg: yeah, legit, the more I use `forge` the more frustrated I am with how slow it is

service_contracts/Makefile
rvagg: mm, fair, done

wjmelements (approved):

PR #527, opened 2026-06-15
feat: share CDN bandwidth rail across a payer's data sets

juliangruber:
## What

Lets a client buy the CDN bandwidth service once and share it across multiple data sets, instead of paying for it per data set. This is the core, on-chain part of a cross-repo proposal.

Full proposal, with rationale and the companion changes in the other repos: https://gist.github.com/juliangruber/a34b225f9588ec68d069054d731e20c9

The gist also links to the companion draft PRs in `FilOzone/synapse-sdk`, `filbeam/contracts`, and `filbeam/worker`.

## Why

The SDK now does multi-copy upload, a piece is stored in 2 data sets on 2 providers. When `withCDN` is enabled on both, each data set creates its own CDN bandwidth rail to the same FilBeam beneficiary, so the bandwidth service is bought twice: the 0.7 USDFC lockup is reserved twice and there are two rails for FilBeam to meter against. The bytes are delivered to the end user once, so this duplication is pure cost to the client.

The cache-miss rail is different, its payee is the data set's SP, which differs across the two copies, so it genuinely stays per provider.

## How

A CDN subscription is keyed by `(payer, groupId)`, where `groupId` is an optional value carried in the `withCDN` metadata entry (its presence was the only thing read before, the value was unused). FWSS maps that key to a single shared bandwidth rail.

- New storage `cdnGroupRail` (subscription key to shared `cdnRailId`) and `cdnRailRefCount` (members per shared rail), appended at slots 23 and 24.
- `createRails` reuses an active shared bandwidth rail when the group key matches, skipping the second rail and its lockup. The first member creates and seeds it. The cache-miss rail is always created per data set.
- `validatePayerOperatorApprovalAndFunds` only requires the bandwidth lockup for the first member of a subscription.
- New `settleCDNBandwidthRail(cdnRailId, cdnAmount)` settles the shared rail once for the whole subscription. `settleFilBeamPaymentRails` remains for cache-miss and legacy ungrouped data sets.
- Teardown (`terminateCDNService`, `dataSetDeleted`, abandonment) decrements the ref count and terminates the bandwidth rail only when the last member leaves.

The shared `cdnRailId` is the subscription identity, so downstream systems aggregate bandwidth by rail id and need no knowledge of group ids.

## Compatibility

Purely additive. An empty `withCDN` value keeps the legacy one-rail-per-data-set behavior, ref counting only engages for grouped data sets. No change to the immutable FilBeam beneficiary and no new signed-message format, the group id is already covered by the EIP-712 creation signature.

## Follow-ups

These are intentionally out of scope for this draft.

- Regenerate the storage-layout JSON and ABIs via the Makefile (the `.sol` layout binding is updated here, the generated `.json` artifact is not).
- Tests for ref-count correctness across the normal-deletion and abandonment teardown paths, and for two data sets in one group sharing a single bandwidth rail.
- A view getter for `cdnRailRefCount` / `cdnGroupRail` if FilBeam wants to read subscription membership on-chain.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -257,6 +257,12 @@ Datasets with CDN support have three payment rails: a **PDP rail** for storage p
 
 Both CDN rails have `paymentRate = 0` and use fixed lockup for one-time payments based on usage. At dataset creation the cache-miss rail is seeded with **0.3 USDFC** and the CDN rail with **0.7 USDFC**. Both CDN rails use a **5-day lockup period**, which sets the settle window FilBeam has after dataset deletion to claim any remaining fixed lockup.
 
+### Shared bandwidth rail (CDN subscriptions)
+
+The bandwidth rail can be shared across multiple data sets of the same payer so CDN is bought once even when a piece is stored in several data sets (for example multi-copy upload across providers). The `withCDN` metadata value carries an optional group id, and FWSS keys a shared bandwidth rail by `keccak256(abi.encode(payer, groupId))`. When a data set is created with a group id whose shared bandwidth rail already exists and is active, the data set joins it, no second bandwidth rail or 0.7 USDFC lockup is created. The shared `cdnRailId` is the subscription identity: every member data set resolves to the same rail, so the FilBeam controller meters and settles bandwidth once per rail via `settleCDNBandwidthRail(cdnRailId, cdnAmount)`.
+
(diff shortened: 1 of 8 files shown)


BigLep:
Marked as potential for https://github.com/FilOzone/filecoin-services/issues/534

Issue #439, opened 2026-03-09
Modular tooling approach for service_contracts operational scripts


rjan90:
## Context

We have a growing set of operational scripts under `service_contracts/tools` for deployment, upgrades, and contract operations. They are currently Bash-heavy, which has been effective for getting workflows in place quickly, but it is starting to make reuse, composition, and testing harder.

## Problem

As the number of scripts grows, we are accumulating:
- duplicated logic around chain selection, deployment address loading, and Safe interactions
- inconsistent operator UX across scripts
- limited structure for shared modules and tests
- more Bash to maintain over time

## Goal

Evaluate whether we should keep building on the current Bash approach, or move some or all of this tooling to a more modular implementation approach.

## Things to explore

- whether there is enough common logic now to justify a shared toolchain
- whether a different language/runtime would improve modularity and code sharing
- whether we should standardize a common CLI/operator interface across scripts
- whether we should add shell tooling now even if we keep Bash for the near term

## Near-term improvements to consider even if we keep Bash

- add `shfmt`
- add `shellcheck`
- document conventions for script inputs/outputs
- identify common helpers that should be consolidated

## Outcome

A recommendation on:
1. stay with Bash + add better tooling, or
2. migrate future operational tooling to a more structured toolchain

rjan90:
Although not in the modular tooling approach direction, I'm removing some stale service contract tools here: https://github.com/FilOzone/filecoin-services/pull/538 to clean things up a little bit

PR #530, opened 2026-06-16
feat: deployment bytecode metadata

### LINKED ISSUE #524: Add initcode hash metadata to deployments.json
If we include initcode hash, constructor args, and library info to deployments.json, it will be able to detect which contracts need to be upgraded.

This idea was mentioned in the original spec of #354:

> The json file might also include other metadata like the commit from which the code was deployed. We could use this metadata to check if the deployed bytecode for the contract has changed since that version.

wjmelements:
Closes #524
Reviewer @zenground0
CC @rjan90
After this change, we no-longer have to manually determine which contracts need to be redeployed.
A contract needs to be redeployed if its initcode (which includes deployed bytecode, constructor args, library addresses) has changed.
There is also a `pinned` flag which can be used to mark contracts as not to be redeployed; for example we do not want to redeploy the proxies.
Not having a contract address in the deployments.json will still force a deployment.
I also add a script that can be used to verify a deployment; I used some of its logic when backfilling these entries.
Once this is merged I will update the deployment runbooks.
#### Test Plan
I ran the deploy all script on testnet and it redeployed Rails library and FWSS as expected, updating deployments.json.
#### Changes
* `needs_deployment`
* `update_deployment_bytecode`
* `verify-deployments.sh`
* backfill previous deployment metadata for mainnet and calibration


service_contracts/tools/deployments.sh
```diff
@@ -187,6 +187,229 @@ update_deployment_metadata() {
...
+       --argjson args "$(_build_args_json "${constructor_args[@]+"${constructor_args[@]}"}")" \
+       'if .[$chain] then . else .[$chain] = {} end |
+        .[$chain].contracts = (.[$chain].contracts // {}) |
+        .[$chain].contracts[$key] = {
+            "initcode_hash": $hash,
+            "artifact_contract": $artifact,
+            "libraries": $libs,
+            "constructor_args": $args
+        }' \
+       "$DEPLOYMENTS_JSON_PATH" > "$temp_file"
```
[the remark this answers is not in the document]
wjmelements: `update_deployment_bytecode` is never called for pinned contracts

service_contracts/tools/deployments.sh
```diff
@@ -187,6 +187,229 @@ update_deployment_metadata() {
...
+    local libraries_str="${2:-}"
+    local initcode_hex
+    initcode_hex=$(jq -r '.bytecode.object' "$artifact_path" | sed 's/^0x//')
+    if [ -n "$libraries_str" ]; then
+        initcode_hex=$(python3 "$SCRIPT_DIR/bytecode.py" link \
+            "$initcode_hex" \
+            "$(jq -c '.bytecode.linkReferences // {}' "$artifact_path")" \
+            "$(_build_libs_json "$libraries_str")")
+    fi
+    printf '%s' "$initcode_hex" | cast keccak
```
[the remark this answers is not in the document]
wjmelements: darn, I have to redo all of the backfilled initcode hashes

service_contracts/deployments.json
```diff
@@ -37,6 +142,110 @@
...
+        },
+        "constructor_args": [
+          "0x85e366Cf9DD2c0aE37E963d9556F5f4718d6417C",
+          "0x09a0fDc2723fAd1A7b8e3e00eE5DF73841df55a0",
+          "0x1D60d2F5960Af6341e842C539985FA297E10d6eA",
+          "0xb3042734b608a1B16e9e86B374A3f3e389B4cDf0",
+          "0x839e5c9988e4e9977d40708d0094103c0839Ac9D",
+          "0x518411c2062E119Aaf7A8B12A2eDf9a939347655",
+          "8"
+        ]
```
rjan90, line 203: I think this will make calibnet FWSS look changed even when it is not. The stored constructor args have the FilBeam beneficiary address before the USDFC token address, but the contract constructor and deploy script expect USDFC first, then FilBeam beneficiary. Because `needs_deployment` compares against this metadata, it will see the correct deploy-script args as different from the backfilled args, and `verify-deployments.sh` will also simulate with the wrong constructor ordering.

service_contracts/tools/warm-storage-deploy-all.sh
rjan90, line 402: I think this will make the existing reinitializer-counter call sites behave differently now that `deploy_implementation_if_needed` uses `needs_deployment` instead of only checking whether an address is already set. These call sites were already passing `proxy counter + 1`, which was fine when the value was only used for an actual deployment, but now it is also used for the metadata comparison. Since the stored metadata has the previous deployed counter, unchanged implementations will look like their constructor args changed and can be redeployed unnecessarily.

service_contracts/deployments.json
wjmelements: If the args are in the wrong order, that is a huge problem. There are bigger reasons that confusing these two parameters would be bad than that the contract would redeploy.

service_contracts/tools/warm-storage-deploy-all.sh
wjmelements: Yes we need to check needs deployment before incrementing the counter

service_contracts/deployments.json
wjmelements: This is a good find. It is fixed in 44afb112c8719794c30025a85261f4ad58d76156.

service_contracts/tools/deployments.sh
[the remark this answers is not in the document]
wjmelements: 44afb112c8719794c30025a85261f4ad58d76156

service_contracts/tools/warm-storage-deploy-all.sh
wjmelements: [a165cc5](https://github.com/FilOzone/filecoin-services/pull/530/commits/a165cc5830c9f4dd196f73f498f83e6b4ab72cf5)

service_contracts/tools/warm-storage-deploy-implementation.sh
```diff
@@ -129,49 +130,86 @@ else
...
+# on deployment, so passing the current (stored+1) value would produce a false "args
+# changed" result when nothing else has changed.
+# Look up the counter's position by its constructor parameter name so a future
+# reordering of the FWSS constructor args doesn't silently read the wrong stored value.
+FWSS_ARTIFACT_PATH=$(_artifact_path "src/FilecoinWarmStorageService.sol:FilecoinWarmStorageService")
+REINIT_ARG_INDEX=$(jq -r '[.abi[] | select(.type=="constructor") | .inputs[].name] | index("_reinitializer_version")' \
+    "$FWSS_ARTIFACT_PATH" 2>/dev/null)
+STORED_FWSS_COUNTER=$(jq -r --argjson idx "${REINIT_ARG_INDEX:-6}" \
+    ".[\"$CHAIN\"].contracts.FWSS_IMPLEMENTATION.constructor_args[\$idx] // empty" \
+    "$DEPLOYMENTS_JSON_PATH" 2>/dev/null)
```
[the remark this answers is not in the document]
wjmelements: a165cc5

service_contracts/tools/verify-deployments.sh
```diff
@@ -0,0 +1,282 @@
...
+                printf '  %-45s SKIP (snapshot artifact; source diverged from deployment)\n' "$contract_key"
+                continue
+                ;;
+        esac
+
+        # Rebuild "path:Name:addr,..." from stored {"path:Name": "addr"} JSON
+        libraries_str=$(jq -r \
+            ".[\"$CHAIN\"].contracts[\"$contract_key\"].libraries | to_entries[] | \"\(.key):\(.value)\"" \
+            "$DEPLOYMENTS_JSON_PATH" 2>/dev/null \
+            | tr '\n' ',' | sed 's/,$//')
```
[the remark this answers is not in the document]
wjmelements: d2cae3ce6aea0a289558920d93d8f0a45c342f01

BigLep:
@wjmelements first off, thank you for the hard work driving this. The many commits over the last week make it clear how much subtle EVM byte-level logic (initcode hashing, CBOR stripping, library linking, via_ir fallbacks) had to be nailed down to get redeployment decisions to be deterministic. Thank you.

I haven't gone deep on the PR itself. This is a meta/passing comment, not a code review.

What I want to flag is our widening bash footprint. By my count `service_contracts/tools/` is already ~4,235 LOC across 24 `.sh` files, with several scripts over 500 LOC each (`warm-storage-manage-approved-provider.sh` ~711, `warm-storage-deploy-all.sh` ~598). This PR adds roughly ~660 more bash LOC (`verify-deployments.sh` +282, `deployments.sh` +345) and also introduces `bytecode.py`, I believe our first Python file in this tree, presumably because some of the new logic just didn't fit in bash. That last point is the real signal for me: we're now implicitly admitting bash has hit its ceiling for this kind of work.

We opened #439 a few months ago asking exactly this question, and it's been sitting open. I think now, coming out of GA, before we accumulate more deployment, upgrade, and verification tooling, is the right moment to actually answer it, rather than continue widening the gap.

Concrete proposal: rather than introducing Python here, would it make sense to adopt **TypeScript** for new operational tooling? It would match what we're already doing for the rest of our contract-related tooling (Synapse SDK, viem-based code), and `execa` keeps the option of shelling out to `forge`/`cast` where that's still the right call. I'm open to other suggestions: the key ask is that we pick one non-bash language and commit, rather than accumulating both Python and bash on top of what's already there. I'm not suggesting we rewrite the existing bash, but rather that new scripts go in the chosen language and existing ones migrate piecewise when touched.

Happy to move this conversation back to #439 if that's a better venue. I don't want to block this PR, but I wanted to raise the concern visibly before we land another ~1000 LOC of Bash and mixed-language tooling.

rjan90 (approved):
Approving. This gets us deterministic deployment metadata and the checks are green.

I agree with @BigLep tooling concern around the growing bash/Python footprint, but I don’t think it should block this PR. Let’s move that decision back to #439 and follow up with a clear direction for new operational tooling

BigLep:
> I agree with @BigLep tooling concern around the growing bash/Python footprint, but I don’t think it should block this PR. Let’s move that decision back to #439 and follow up with a clear direction for new operational tooling

My biggest concerns with moving forward are: 
1. We lose the forcing function to start doing the refactor/cleanup. (It's hard to prioritize a cleanup/improvement when it's not part of something delivering immediate user value)
2. We're introducing a new language here (Python in this case) to do some special functionality that was tougher to do in shell scripts and we don't have any testing around it. If we move to a typescript world, it would be ideal to have some test cases around this Python code we can easily verify we haven't screwed things up in the Python to typescript migration. (I haven't looked closely enough at the Python, maybe that isn't a concern?)

Anyways, I won't hold this up as it's hard to justify spending more time on it now. If some test/verification cases can be easily added, that's great. Otherwise we merge.

wjmelements:
I think that if we want to change the scripts to ts, that should be a big separate PR, and I think AI can do it.

rjan90:
@wjmelements is this fine to merge now? Or are there specific parts that you want @ZenGround0 to also review?

PR #536, opened 2026-06-23
Optional dataset-level programmable acls

essamhassan:
## Summary

Optional per-dataset `IDataSetAuthorizer` write-authorization hook. No authorizer → existing payer/session-key behavior unchanged. Authorizer set → it is the sole gate for add-pieces, schedule-removals, and signed termination.

## Behavior

- Payer sets/rotates/clears the authorizer; state view exposes it; dataset deletion clears it.
- Receives `dataSetId`, `payer`, operation typehash, EIP-712 digest, raw signature, and ABI-encoded `operationData`.
- `isAuthorized` is a state-mutating CALL (authorizers may keep nonces/rate-limits): `false` → `Errors.Unauthorized`; a revert bubbles.
- Griefing/reentrancy: 150M gas cap (enforcement verified live on FEVM) + an EIP-1153 transient reentrancy latch scoped to the authorization path.

## Tests

Backward compat; delegated add/removal/termination; session keys via owner; `operationData` forwarding; real `vm.sign` recovery; false-vs-revert; state mutation during authorization; reentrancy latch (blocks + clears); cleanup on deletion.


service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,16 @@
...
+///         FWSS accepts the default payer / session-key signature, and otherwise
+///         asks the authorizer whether the recovered signer is delegated for the
+///         operation — letting one payer host many data sets with distinct ACLs.
+interface IDataSetAuthorizer {
+    /// @param dataSetId The data set being operated on.
+    /// @param signer The address recovered from the operation signature.
+    /// @param operation The operation type hash. Consumers can read stable values
+    ///        from FWSS, e.g. ADD_PIECES_OPERATION().
+    /// @return True if `signer` is allowed to perform `operation` on `dataSetId`.
+    function isAuthorized(uint256 dataSetId, address signer, bytes32 operation) external view returns (bool);
```
essamhassan: I would pass the metadata here as well, to make sure we can give granular access not just dataset level based on passed signed metadata

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1432,34 +1448,80 @@ contract FilecoinWarmStorageService is
...
+        signer = SignatureVerificationLib.recoverSigner(digest, signature);
+        require(signer != address(0), Errors.InvalidSignature(payer, signer));
+
+        if (signer == payer || sessionKeyRegistry.authorizationExpiry(payer, signer, operation) >= block.timestamp) {
+            return signer;
+        }
+
+        address authorizer = dataSetAuthorizer[dataSetId];
+        require(
+            authorizer != address(0) && IDataSetAuthorizer(authorizer).isAuthorized(dataSetId, signer, operation),
```
essamhassan: this should be called with gas budget to make sure generic impl is capped and protect SP from unpredictable gas cost

service_contracts/src/interfaces/IDataSetAuthorizer.sol
wjmelements: to allow non-secp256k1 signers, you would want to consider alternatives to `address signer`. Is secp256r1 in scope? Should we perhaps pass the entire signature along with the hash?

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -81,12 +81,18 @@ contract FilecoinWarmStorageService is
...
+    // intentionally excluded: creating a data set is the payer's own act (it sets
+    // the payer and binds the rails), so it is never delegated to an authorizer.
     bytes32 public constant ADD_PIECES_OPERATION = SignatureVerificationLib.ADD_PIECES_TYPEHASH;
     bytes32 public constant SCHEDULE_PIECE_REMOVALS_OPERATION =
         SignatureVerificationLib.SCHEDULE_PIECE_REMOVALS_TYPEHASH;
     bytes32 public constant TERMINATE_SERVICE_OPERATION = SignatureVerificationLib.TERMINATE_SERVICE_TYPEHASH;
 
+    // Gas budget for the optional authorizer staticcall, so a generic authorizer
+    // cannot impose unpredictable gas cost on the SP submitting the write.
+    uint256 private constant MAX_AUTHORIZER_GAS = 100_000;
```
wjmelements: I don't think this works on filecoin. I think all gas is always forwarded, and also that 100k is not very much gas on filecoin.

essamhassan: 100k was representative tbh it should be much larger
re gas:xxx not working on FEVM, how do you recommend we do the capping? (this can be done with gas limits on the SPs but harder to attribute to auth)

service_contracts/src/interfaces/IDataSetAuthorizer.sol
essamhassan: we can pass signature/hash for optionality downstream but wouldn't non secp256k1 already fail the upstream recover_signer?

service_contracts/src/FilecoinWarmStorageService.sol
essamhassan: got an LLM assisted probe contract to test gas budgets on FEVM 0x2239129799B6581c9eD47ED3A733b7aE083ACF7D

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

contract GasProbe {
    event ProbeResult(uint256 forwarded, uint256 consumedByCall, bool caught);

    // Unbounded gas sink (never returns); consumes whatever gas it is given.
    function grief() external pure {
        uint256 x;
        for (uint256 i = 0;; i++) {
            x = uint256(keccak256(abi.encodePacked(x, i)));
        }
    }

    // Calls grief with an explicit {gas: gasLimit}. If FEVM honors the stipend,
    // consumedByCall ~= gasLimit. If "all gas is forwarded", consumedByCall ~= 63/64
    // of the whole tx (>> gasLimit).
    function probe(uint256 gasLimit) external {
        uint256 g0 = gasleft();
        bool caught;
        try this.grief{gas: gasLimit}() {} catch { caught = true; }
        emit ProbeResult(gasLimit, g0 - gasleft(), caught);
    }
}
```

it ran few iterations to test gas budgets downstream against a 1B tx level gas limit

Nr1
block: 3836236
tx hash: 0xfeaa6edda21f654ab6789243d8a1b1fb8a89fe8581bde0a53d1636f539335f4c
what: probe(10_000_000) (first send — output mis-parsed locally but landed on-chain)
result: forwarded 10M → consumed ~11.34M, caught
────────────────────────────────────────
Nr2
block: 3836239
tx hash: 0x4613b4497984b2ab0e924b5ca2d073019bd65924f84f15493f901a52a76f9651
what: probe(10_000_000) (re-run)
result: forwarded 10M → consumed 11,335,696, caught · tx gas 13,677,939
────────────────────────────────────────
Nr3
block: 3836242
tx hash: 0xdbb6f26be433679af57c06b32b3ff6181711d4bace2f7fb3a79d853d2d0cb31c
what: probe(50_000_000)
result: forwarded 50M → consumed 51,335,696, caught · tx gas 53,677,939

service_contracts/src/interfaces/IDataSetAuthorizer.sol
wjmelements: pass the whole sig and do the ecrecover yourself if you're setting this override

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1415,51 +1432,91 @@ contract FilecoinWarmStorageService is
...
-        SignatureVerificationLib.verifyAddPiecesSignature(payer, signature, digest, sessionKeyRegistry);
+        SignatureVerificationLib.verifySignatureWithAuthorizer(
+            payer,
+            signature,
+            digest,
+            sessionKeyRegistry,
+            ADD_PIECES_OPERATION,
+            dataSetId,
+            dataSetAuthorizer[dataSetId],
+            MAX_AUTHORIZER_GAS,
```
wjmelements: I don't think this should be a parameter if it is a constant

service_contracts/src/lib/SignatureVerificationLib.sol
```diff
@@ -290,4 +291,32 @@ library SignatureVerificationLib {
...
+    ) public view returns (address signer) {
+        signer = recoverSigner(digest, signature);
+        require(signer != address(0), Errors.InvalidSignature(payer, signer));
+
+        if (signer == payer || sessionKeyRegistry.authorizationExpiry(payer, signer, operation) >= block.timestamp) {
+            return signer;
+        }
+
+        require(authorizer != address(0), Errors.InvalidSignature(payer, signer));
+        try IDataSetAuthorizer(authorizer).isAuthorized{gas: authorizerGas}(
```
wjmelements: I don't think we need to worry about gas griefing. All this does is allow a user to restrict their own ability to perform consensual actions with the SP, and the SP should still reject such actions if the gas cost is greater than the operation fee. So the SP is already protected (by operational fees (#469)), and the user is still paying for it.

service_contracts/src/lib/SignatureVerificationLib.sol
```diff
@@ -290,4 +291,32 @@ library SignatureVerificationLib {
...
+        bytes calldata signature,
+        bytes32 digest,
+        SessionKeyRegistry sessionKeyRegistry,
+        bytes32 operation,
+        uint256 dataSetId,
+        address authorizer,
+        uint256 authorizerGas,
+        bytes calldata metadata
+    ) public view returns (address signer) {
+        signer = recoverSigner(digest, signature);
```
wjmelements: don't recover signer, and don't look at session key registry. if there is union, do it in the IDataSetAuthorizer

service_contracts/src/lib/SignatureVerificationLib.sol
```diff
@@ -290,4 +291,32 @@ library SignatureVerificationLib {
...
+        }
+
+        require(authorizer != address(0), Errors.InvalidSignature(payer, signer));
+        try IDataSetAuthorizer(authorizer).isAuthorized{gas: authorizerGas}(
+            dataSetId, signer, operation, digest, signature, metadata
+        ) returns (bool ok) {
+            require(ok, Errors.InvalidSignature(payer, signer));
+        } catch {
+            revert Errors.InvalidSignature(payer, signer);
+        }
```
wjmelements: don't catch

service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+pragma solidity ^0.8.20;
+
+/// @title IDataSetAuthorizer
+/// @notice Optional per-data-set write ACL.
+/// @dev FWSS calls this with a fixed gas budget. Revert or out-of-gas means "not authorized".
+interface IDataSetAuthorizer {
+    /// @param dataSetId The data set being operated on.
+    /// @param signer The secp256k1 signer recovered from the operation signature.
```
wjmelements: I think we should drop the signer from this and let the authorizer compute this information.

service_contracts/src/lib/SignatureVerificationLib.sol
```diff
@@ -290,4 +291,32 @@ library SignatureVerificationLib {
...
+
+        if (signer == payer || sessionKeyRegistry.authorizationExpiry(payer, signer, operation) >= block.timestamp) {
+            return signer;
+        }
+
+        require(authorizer != address(0), Errors.InvalidSignature(payer, signer));
+        try IDataSetAuthorizer(authorizer).isAuthorized{gas: authorizerGas}(
+            dataSetId, signer, operation, digest, signature, metadata
+        ) returns (bool ok) {
+            require(ok, Errors.InvalidSignature(payer, signer));
```
wjmelements: can be `error Unauthorized(payer, operation, digest, signature)`

service_contracts/tools/generate_storage_layout.sh
```diff
@@ -8,7 +8,7 @@ echo // This file is a generated binding and any changes will be lost.
 echo // Generated with tools/generate_storage_layout.sh
 echo
 
-forge inspect --json $1 storageLayout \
```
wjmelements, line 11: why do you remove the `--json`?

service_contracts/Makefile
```diff
@@ -34,11 +34,11 @@ build:
 
 # Storage layout generation
 $(LAYOUT): tools/generate_storage_layout.sh src/FilecoinWarmStorageService.sol
-	$^ | forge fmt -r - > $@
+	tools/generate_storage_layout.sh src/FilecoinWarmStorageService.sol:FilecoinWarmStorageService | forge fmt -r - > $@
```
wjmelements: This is `$^`. Why did you change this?

service_contracts/Makefile
```diff
@@ -34,11 +34,11 @@ build:
...
 
 # Storage layout JSON (full metadata for upgrade safety checks)
 $(LAYOUT_JSON): src/FilecoinWarmStorageService.sol $(LAYOUT_SNAPSHOT_JQ)
-	forge inspect --json $< storageLayout | jq -f $(LAYOUT_SNAPSHOT_JQ) > $@
```
wjmelements, line 41: why do you remove `--json`?

service_contracts/Makefile
```diff
@@ -34,11 +34,11 @@ build:
...
+	forge inspect src/FilecoinWarmStorageService.sol:FilecoinWarmStorageService storageLayout | jq -f $(LAYOUT_SNAPSHOT_JQ) > $@
```
wjmelements: I think you might be running the wrong version of foundry. Downgrade to 1.3.5. We will not be upgrading anytime soon.

service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,23 @@
...
+    /// @param operation The operation type hash, e.g. FWSS.ADD_PIECES_OPERATION().
+    /// @param digest The EIP-712 digest signed for the operation.
+    /// @param signature The raw signature over `digest`.
+    /// @param metadata ABI-encoded signed operation payload forwarded by FWSS.
+    /// @return True if `signer` is allowed to perform `operation` on `dataSetId`.
+    function isAuthorized(
```
wjmelements, line 23: I think it should also pass payer

service_contracts/src/lib/SignatureVerificationLib.sol
essamhassan: changed to only recover signer if authorizer not set

service_contracts/src/lib/SignatureVerificationLib.sol
essamhassan: removed budgets

wjmelements:
One kind of griefing that would be hard to detect is a randomized authorizer. Suppose the validator approved 1% of signatures (regardless of what they were). There would be a 1% chance they would pass `eth_estimateGas` (and the griefer would only submit ones that succeed), but then have a 99% chance to fail when the transaction is submitted, thereby wasting a lot of resources. In fact you could construct a signature that would succeed in one block but fail in all subsequent ones. I don't know a good way to detect or mitigate this griefing (because it is indistinguishable from an SP submitting with fake signatures), so we would probably have to have a whitelist for authorizers, on or off chain.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -80,6 +80,11 @@ contract FilecoinWarmStorageService is
     // Version tracking
     string public constant VERSION = "1.3.0";
 
+    bytes32 public constant ADD_PIECES_OPERATION = SignatureVerificationLib.ADD_PIECES_TYPEHASH;
+    bytes32 public constant SCHEDULE_PIECE_REMOVALS_OPERATION =
+        SignatureVerificationLib.SCHEDULE_PIECE_REMOVALS_TYPEHASH;
+    bytes32 public constant TERMINATE_SERVICE_OPERATION = SignatureVerificationLib.TERMINATE_SERVICE_TYPEHASH;
```
wjmelements: I have a preference against adding public constants. I don't even like the ones we already have. The main reason is codesize. Another reason is that solidity's selector dispatch is pretty inefficient, so this produces overhead affecting every other method. If the code gets large enough, we have to build with `runs = 0` to minimize the codesize, and then solidity will use a linear dispatcher.

Do you need to surface these constants? Can you just import them? They aren't going to change.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -303,8 +308,13 @@ contract FilecoinWarmStorageService is
     // Piece IDs awaiting metadata cleanup; cleared each nextProvingPeriod call
     mapping(uint256 dataSetId => uint256[] pieceIds) internal scheduledPieceMetadataRemovals;
 
+    // Optional per-data-set authorizer (address(0) = default payer/session-key behavior).
+    mapping(uint256 dataSetId => address authorizer) internal dataSetAuthorizer;
```
wjmelements, line 314: I like that the mapping is internal.

Add view methods for this to the FilecoinWarmStorageServiceStateLibrary and then `make gen`. Then you can write an integration test with your test authorizer that can check that `dataSetAuthorizer` was set using the View contract.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: Also it seems that with `--via-ir`, it will always use a linear dispatcher (!).

service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,25 @@
...
+///         delegates the entire authorization decision for that data set to it.
+/// @dev A revert (or out-of-gas) means "not authorized": the operation reverts.
+interface IDataSetAuthorizer {
+    /// @param dataSetId The data set being operated on.
+    /// @param payer The data set's payer (the on-chain owner of the rails).
+    /// @param operation The operation type hash, e.g. FWSS.ADD_PIECES_OPERATION().
+    /// @param digest The EIP-712 digest signed for the operation.
+    /// @param signature The raw signature over `digest`; the authorizer recovers the
+    ///        signer itself, on whatever curve it supports.
+    /// @param metadata ABI-encoded signed operation payload forwarded by FWSS.
```
wjmelements: I am suspicious of this `metadata` encoding. Perhaps I don't know what it is for or why it is necessary. It might need a better name since what we call metadata is file and dataset metadata such as IPNI info.

Is this always going to the be the signed payload (are we allowing alternative hash functions or alternative payloads)? If so perhaps we don't need to pass the digest, and perhaps this could be named something like "signedData". Perhaps we should pass these fields individually and let the authorizer contract calculate what they want.

keccak is a good hash function and I wouldn't mind if we keep enshrining it.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1415,51 +1429,88 @@ contract FilecoinWarmStorageService is
...
+        require(authorizer == address(0) || authorizer.code.length > 0, Errors.InvalidDataSetAuthorizer(authorizer));
+        dataSetAuthorizer[dataSetId] = authorizer;
+        emit DataSetAuthorizerSet(dataSetId, authorizer);
+    }
+
+    /**
+     * @notice The authorizer attached to a data set, or address(0) if none.
+     */
+    function getDataSetAuthorizer(uint256 dataSetId) external view returns (address) {
+        return dataSetAuthorizer[dataSetId];
```
wjmelements: Remove. Use `Extsload` pattern for state view.

* https://github.com/FilOzone/filecoin-services/pull/536/changes#r3493987734
* https://github.com/FilOzone/filecoin-services/pull/536/changes#r3493961494

wjmelements (changes requested):
Finished today's pass. This is pretty close. The most important thing to decide in this PR is the IDataSetAuthorizer interface.

I have solicited review from some other team members. I hope they can review this before next week so we can try to merge it then.

service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,25 @@
...
+pragma solidity ^0.8.20;
+
+/// @title IDataSetAuthorizer
+/// @notice Optional per-data-set write ACL. When a payer attaches an authorizer, FWSS
+///         delegates the entire authorization decision for that data set to it.
+/// @dev A revert (or out-of-gas) means "not authorized": the operation reverts.
+interface IDataSetAuthorizer {
+    /// @param dataSetId The data set being operated on.
+    /// @param payer The data set's payer (the on-chain owner of the rails).
+    /// @param operation The operation type hash, e.g. FWSS.ADD_PIECES_OPERATION().
```
snadrus: seems like a short enum

service_contracts/src/FilecoinWarmStorageService.sol
essamhassan: removed all constants

service_contracts/src/FilecoinWarmStorageService.sol
essamhassan: removed, moved to Extsload https://github.com/FilOzone/filecoin-services/pull/536/changes#diff-abd1e87020595cc760e5135e4adc1fea1141fd8d02a335864e9f17c7a9bd4d63R234

service_contracts/src/FilecoinWarmStorageService.sol
essamhassan: done

service_contracts/src/interfaces/IDataSetAuthorizer.sol
essamhassan: renamed it to signedData, the thinking here is we want to pass extra information in the signedData to allow fine-grained access based on the fields provided

essamhassan: @wjmelements happy to discuss this

service_contracts/src/interfaces/IDataSetAuthorizer.sol
essamhassan: @snadrus I would not use an enum here just because the usage of these ops downstream might rely on their values upstream.
 example: a non upgradable Authorizer that is using enums and then retranslating back to signature to access SKR would break if upstream signatures were upgraded on FWSS.
happy to go the enum route if we can preserve that functionality

essamhassan: @wjmelements if these constants are importable in the authorizer, we can move to enum and let users import them when needed if they want to do the enum=>type hash mapping?

service_contracts/src/interfaces/IDataSetAuthorizer.sol
wjmelements: signedData is fine because the data is operation-specific

service_contracts/src/interfaces/IDataSetAuthorizer.sol
wjmelements: I don't want to use an enum for the operations. We decided against enumerations early on because the SessionKeyRegistry is global and not necessarily specific to FWSS.

Previously we also considered fixed32 strings, which has some benefit of clarity over a hash, but the wallets didn't render it in a friendly way either way and bytes32 was simpler.

service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,25 @@
...
+    /// @param signedData The ABI-encoded signed operation payload forwarded by FWSS.
+    /// @return True if the operation is authorized on `dataSetId`.
+    function isAuthorized(
+        uint256 dataSetId,
+        address payer,
+        bytes32 operation,
+        bytes32 digest,
+        bytes calldata signature,
+        bytes calldata signedData
+    ) external view returns (bool);
```
wjmelements: outvars can also have names, like `returns (bool authorized)`

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1415,51 +1424,86 @@ contract FilecoinWarmStorageService is
...
-        SignatureVerificationLib.verifyAddPiecesSignature(payer, signature, digest, sessionKeyRegistry);
+        SignatureVerificationLib.verifySignatureWithAuthorizer(
+            payer,
+            signature,
+            digest,
+            sessionKeyRegistry,
+            SignatureVerificationLib.ADD_PIECES_TYPEHASH,
+            dataSetId,
+            dataSetAuthorizer[dataSetId],
+            abi.encode(clientDataSetId, nonce, pieceDataArray, allKeys, allValues)
```
wjmelements: This duplicates much of the work of addPiecesStructHash. So let's remove such methods so that we can avoid recalculating `abi.encode` for these operations.

essamhassan: In addPieces the data passed are hashed here https://github.com/essamhassan/filecoin-services/blob/8606c72953d08b1d1e349dddb903605921f05516/service_contracts/src/lib/SignatureVerificationLib.sol#L88-L104 and we would like authorizer to have access to the raw data. seems this apply to removals as well with `keccak(pieceIds)`

service_contracts/src/interfaces/IDataSetAuthorizer.sol
essamhassan: done

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: yeah, I think there are four operations right now

essamhassan: what we can do is move the digest calc in the call and only precompute digest for the backward compat flow. In the authorizer flow, give authorizer the optionality to compute digest or not. this means strictly 1 encode / call unless authorizer want to compute the digest from params (which the authorizer already have all of them)

snadrus (approved):

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: Yes, we don't want to duplicate keccak or abi.encode. It is okay to remove those methods from the signature library.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -303,8 +303,13 @@ contract FilecoinWarmStorageService is
     // Piece IDs awaiting metadata cleanup; cleared each nextProvingPeriod call
     mapping(uint256 dataSetId => uint256[] pieceIds) internal scheduledPieceMetadataRemovals;
 
+    // Optional per-data-set authorizer (address(0) = default payer/session-key behavior).
+    mapping(uint256 dataSetId => address authorizer) internal dataSetAuthorizer;
```
LexLuthr, line 314: Can we also drop this when dataSet is deleted?

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -5262,6 +5393,190 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
...
+        _addAuthorizerTestPiece(dataSetId, 1);
+
+        // The payer is no longer special-cased. Attaching an authorizer delegates every write
+        // decision to it, so a reverting authorizer locks out the payer too.
+        makeSignaturePass(client);
+        vm.expectRevert();
+        _addAuthorizerTestPiece(dataSetId, 2);
+    }
+
+    function testDataSetAuthorizerReceivesSignedDataForEachWrite() public {
```
LexLuthr: This is the strongest FWSS authorizer integration test, but it still relies on makeSignaturePass/ecrecover mocking. Can we add one real vm.sign case through this path, even if only for addPieces, so the FWSS -> authorizer -> recoverSigner flow is covered end-to-end?

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -5262,6 +5393,190 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
...
+        vm.prank(client);
+        pdpServiceWithPayments.setDataSetAuthorizer(dataSetId, address(authorizer));
+
+        Cids.Cid[] memory pieceData = new Cids.Cid[](1);
+        pieceData[0] = Cids.CommPv2FromDigest(0, 4, keccak256(abi.encodePacked("acl_signed_data")));
+        (string[] memory pieceKeys, string[] memory pieceValues) = _getSingleMetadataKV("path", "/acl/piece");
+
+        authorizer.expectAdd(dataSetId, clientDataSetId, 100, keccak256(pieceData[0].data));
+        makeSignaturePass(bob);
+        vm.expectRevert();
```
LexLuthr: Can we make this revert assertion precise? This path is testing an authorizer returning false, not an authorizer revert, so it should assert Errors.Unauthorized(...) rather than generic vm.expectRevert(). That would lock down the intended boundary: false return => FWSS Unauthorized; authorizer revert => bubbled authorizer revert.

LexLuthr (commented):
This is a good solution.
1. We need to update the original PR description as there are some discrepancy between the description and code. 
2. We need to figure out "Authorizer griefing" and trust model for this. It is outside the purview of this PR but it would be good to have a note calling it out in the description before we merge. We should probably also open an issue for it once we merge this PR.

service_contracts/test/FilecoinWarmStorageService.t.sol
essamhassan: addressed here: https://github.com/FilOzone/filecoin-services/pull/536/changes/4c171d73d85822d662d9f8cccdcc0e154bcbbf6a#diff-608583698f2426bc9957d57ef4492accf61e515b4872abaa19b187ffb0fd0535R5591-R5593

service_contracts/src/FilecoinWarmStorageService.sol
essamhassan: addressed here: https://github.com/FilOzone/filecoin-services/pull/536/changes/4c171d73d85822d662d9f8cccdcc0e154bcbbf6a#diff-3c6b40a4c618b442226b6ec0ce0e4937e701306ef505c10918f2e86f4e1fd943R720

essamhassan:
> This is a good solution.
> 
> 1. We need to update the original PR description as there are some discrepancy between the description and code.
> 2. We need to figure out "Authorizer griefing" and trust model for this. It is outside the purview of this PR but it would be good to have a note calling it out in the description before we merge. We should probably also open an issue for it once we merge this PR.

updated PR description and addressed comments. re gas griefing I agree this needs to be addressed added a minimal note in the PR description.

service_contracts/test/FilecoinWarmStorageService.t.sol
essamhassan: addressed here: https://github.com/FilOzone/filecoin-services/pull/536/changes/4c171d73d85822d662d9f8cccdcc0e154bcbbf6a#diff-608583698f2426bc9957d57ef4492accf61e515b4872abaa19b187ffb0fd0535R5439-R5462

essamhassan: @LexLuthr can you confirm this does what you had in mind

LexLuthr: Yes. Thank you.

LexLuthr (approved):
Approving with caveat that changes and issues raised by @wjmelements will be fixed.

wjmelements:
> 2\. We need to figure out "Authorizer griefing" and trust model for this.

I previously recommended a whitelist. But perhaps nondeterministic behavior could be detectable via a trace. If the trace contains timestamp, block number, prevrandao, drand, base fee, gas price, or gas left, then we can't trust `eth_estimateGas`. This approach would be more difficult, and every time someone manages to beat it, we would be paying out bounties. I also am unsure how good structLog traces are in filecoin. This approach might be worth considering if it would be less maintenance than whitelisting.

wjmelements (approved):

rvagg:
A bit too much history in here for me to catch up on but I see that there's no gas cap applied to this any more, why not? That would seem like the most obvious griefing angle in two ways: cost the SP significantly more than we're budgeting to pay them in equivalent USDFC for "average" operations. The gas cost could be elevated quite a bit and the client doesn't feel that pain. The other way is that you could lock up a data set with this if the client can control this authoriser contract. So the SP needs a new failure mode where it stops proving and needs to terminate. Is that an acceptable failure mode?

Also more events in here might be helpful, each time we use the authoriser it would be good to surface what happened in an event.

Lint and gen checks are failing in CI.

wjmelements:
> A bit too much history in here for me to catch up on

Use Claude.

> The other way is that you could lock up a data set with this if the client can control this authoriser contract. So the SP needs a new failure mode where it stops proving and needs to terminate.

No. An authorizer that always rejects is equivalent to a client who walks away. SPs can already terminate the data set unilaterally.

> A bit too much history in here for me to catch up on but I see that there's no gas cap applied to this any more, why not?

Because that doesn't solve the problem at all. Operation fees are not denominated in gas units. To ensure the operation fee pays for the operation, you have to know the gas of the entire operation, not just this subcall. You also have to know the gas price and the price ratio of the payment token to the gas token. This information is all available to the SP when they sign the transaction. I already suggested (on [Github](https://github.com/FilOzone/filecoin-services/pull/536#discussion_r3484094777), on [Slack](https://filecoinproject.slack.com/archives/C08TVNKJV7C/p1782753301729439)) that Curio should reject operations where the operation fee does not cover the cost, or keep them until the gas price comes down (like a mempool).

> most obvious griefing angle

I [describe](https://github.com/FilOzone/filecoin-services/pull/536#issuecomment-4835151143) a much worse griefing strategy. It is worse because the SP does not get paid (which cannot be fixed without also allowing the SP to grief the client). I also described [two](https://github.com/FilOzone/filecoin-services/pull/536#issuecomment-4857643777) ways to solve it. Please review that discussion. I think static analysis is possible and I have done something similar in the past (for ERC20).

rvagg:
> Because that doesn't solve the problem at all. Operation fees are not denominated in gas units. 

It's not about "solving", actually solving for the disconnect between fees paid and costs by the SP isn't something we're attempting to do, we're doing it directionally and adjusting. Over time we'll do an analysis of what it actually cost on average for SPs and adjust pricing accordingly. What I'm suggesting is that a cap is appropriate here to limit the damage and make it so we don't end up adjusting our pricing based on what clients are foisting on SPs. _If_ we want something more unbounded here, then I'd suggest we levy an additional fee for just enabling this feature. Client wants complexity and gas? Fine, pay for it.

essamhassan:
> we're doing it directionally and adjusting. Over time we'll do an analysis of what it actually cost on

@rvagg earlier version had an authorizer gas limit, I don't have particularly strong opinion on this as long as it's generous enough for a sane auth call. early discussions with @wjmelements noted that there is already a hard limit on the cost posed on the SP in the per op cost.

wjmelements:
@rvagg

1. The DoS vector I raised is more significant and it cannot be fixed with a gas limit. Please read it.
2. The solution I propose also solves the issue you raised. Please read it.
3. Therefore, you should catch up on the discussion in this topic before blocking this PR. Please read it.

BigLep:
@wjmelements : I'm not up on the specifics of this PR, and I understand there is discussion getting resolved, but while I'm thinking about it, is there any issue with it rolling out in a FWSS contract upgrade? Does it require special attention, user notification, documentation, Synapse support, etc?

wjmelements:
> I also described [two](https://github.com/FilOzone/filecoin-services/pull/536#issuecomment-4857643777) ways to solve it.

I had some discussion with @jennijuju earlier and we agreed there were some other benefits to the whitelist solution, particularly that we could document the availability of approved authorization registries and how to use them. This whitelist could live on or off-chain: worth discussing.

> is there any issue with it rolling out in a FWSS contract upgrade? Does it require special attention, user notification, documentation, Synapse support, etc?

@BigLep It requires some curio changes (reject or delay over-cost actions, the client-auth whitelist if we do that off-chain), and we should document it and add support to synapse once we have some ACLs whitelisted (probably after deployment), but otherwise doesn't require any immediate special care.

wjmelements:
> This whitelist could live on or off-chain: worth discussing.

The three kinds of authenticator whitelist we could do:
1. **Hard**: enforced in the contract with `require` (checked with `eth_estimateGas`)
2. **Soft**: managed on-chain (checked with `eth_call`) but not enforced in the contract
3. Purely in **Curio**: new auth supported with new curio releases

I don't have a strong preference.

BigLep:
First, genuine thanks to @essamhassan for the effort here, and to everyone who's put review cycles into it. It sees the iteration has been thorough and the `IDataSetAuthorizer` interface is clearly a thoughtful piece of work. Nothing below is a knock on the quality of the contribution.

I'm admittedly no expert here, and there's a lot about our contracts and system I'm still learning. But as I've been thinking about this effort and digging into it more, I have some thoughts I want to share. I realize I may be off base on some of this, and I'm very happy to be corrected.

My instinct is that we should **pause this PR rather than merge it right now**, for a few reasons:

1. **I don't think it has gone through a design phase.** From what I can tell, this is a new and effectively permanent extension point on FWSS, adding contract surface and on-chain state that I understand we can't cleanly migrate away from later. For a change with that kind of lock-in, I'd have expected a short design doc up front laying out the motivation, the alternatives considered, and why this shape is the right one. Right now a lot of the rationale seems to live in resolved review threads, which makes it hard for me (and probably others later) to reconstruct why the design is what it is. If I'm missing where that design thinking happened, please point me to it.

2. **I don't think it's currently a prioritized product need.** As far as I know this wasn't something we set out to build, and it isn't on our roadmap. I know this is a contribution from @essamhassan (thank you), but before we take on the ongoing cost of what looks to me like a new plugin system, I'd want a clearer product signal that this fills a real, prioritized gap. If that signal does exist, I'd love for someone to surface it here so we can justify and plan the work properly.

3. **I think it overlaps with other work we haven't reconciled it against.** In particular the session key registry (to my eye an authorizer looks a bit like a second, differently-shaped session-key mechanism) and the ACL / encryption work @Kubuxu  has been scoping. I know there have also been discussions about secp256r1 / browser-encryption. My worry is that we could end up with several isolated pieces of functionality that we then can't pull together without a larger FWSS revision. I'd feel better looking at these together, possibly including revisiting the session key registry itself, rather than adding another one-off. I could be wrong about how much these actually overlap, so I'd welcome input from people closer to it.

To be clear, I'm suggesting a pause, not a rejection, and I don't want to leave this contribution in limbo. What I think would help is: (a) confirming the product priority, and if it's there, (b) writing up the design so we can evaluate the shape and how it interacts with session keys, ACLs/encryption, and pricing more holistically.

BigLep:
For visibility, we're using slack for quicker/tighter coordination on next steps between Lightshift and FOC working group in a private slack channel: https://filecoinproject.slack.com/archives/C0BE621FD98/p1783511763289569 . We'll report back here with decisions.

wjmelements:
I met with @rvagg. We agreed on the whitelisting approach for DoS. We can develop a static analysis approach if these truly proliferate enough to justify it (but I don't think they will).

jennijuju:
> > I also described [two](https://github.com/FilOzone/filecoin-services/pull/536#issuecomment-4857643777) ways to solve it.
> 
> I had some discussion with @jennijuju earlier and we agreed there were some other benefits to the whitelist solution, particularly that we could document the availability of approved authorization registries and how to use them. This whitelist could live on or off-chain: worth discussing.


I was lack of context and after reading through the full PR/design doc, I don’t think the whitelist mitigation is sufficient here.

It might address protocol security gap but I think the biggest blocker is from the product side: I don’t see how SP gas cost for isAuthorized gets covered.

For context, we went through this exact problem for the base write path #469. The whole reason per-op pricing exists (ADD_PIECES_BASE_FEE, ADD_PIECES_PER_PIECE_FEE, etc.) is so SPs aren’t stuck absorbing unpredictable gas cost — fees are calibrated against known, bounded gas per op and reflected in what the customer actually pays.

The current authorizer design breaks that assumption. isAuthorized can be arbitrarily expensive — P256 verification, UCAN-style path parsing, recursive permission checks, whatever a given authorizer implements,  but the op fee for e.g. addPieces stays fixed and authorizer-agnostic. So either:
1. SPs eat the extra cost on authorizer-enabled datasets, or
2. they build their own per-authorizer cost estimation with no protocol support, or
3. they just decline to service datasets with expensive authorizers.

None of these are good outcomes: (1)/(2) push SP-facing operational cost modeling onto SPs at a stage where that’s not the work they signed up for — they’re here to provide storage service, and from talking to many SPs, understanding chain cost is already a known pain point for biz planning and potential walk-away cause. (3) undercuts adoption for exactly the platforms this feature is meant to serve.

A whitelist might address malicious or nondeterministic cost, but it doesn’t solve the underlying gap: predictable pricing isn’t embedded into the operation/deal pricing itself for a legitimate, well-behaved authorizer.

Context on why this matters: Filecoin SPs are effectively DC operators — they know their infra and ops costs and price accordingly. Onchain gas is outside that comfort zone; it’s an extra cost of supporting on-chain primitives on top of the storage service they actually signed up to provide. That’s exactly why we benchmark gas cost as platform builders and embed it into op pricing in the first place — SPs shouldn’t have to re-derive that per integration, per authorizer.

Is there a plan for authorizer-tiered or authorizer-declared cost (e.g. authorizer publishes expected gas, op fee scales to match)?

jennijuju:
One potential mitigation could be pairing the whitelist with a declared max gas cost for operations like isAuthorized (anything SPs need to pay for), binding it into pricing:
- Declare a max gas cost at whitelist time. When an authorizer gets white listed, require it to declare a benchmarked worst-case gas cost for isAuthorized.
- Priced as a surcharge. That declared max determines an authorizer-specific fee surcharge on top of the base op fee — same mechanism #469 uses for ADD_PIECES_BASE_FEE/ADD_PIECES_PER_PIECE_FEE — so SPs get paid for it.
- Enforced as a hard gas cap. If the call exceeds the declared max, revert.


That said, I’m not actually a fan of this…. It adds real ongoing complexity: every authorizer needs to define its gas cost and pricing correctly, and as maintainers we’d be on the hook for validating those declarations aren’t lowballed — indefinitely, per authorizer, forever. That’s not a one-time review, it’s a standing operational burden.

So stepping back — I think the more useful question isn’t “how do we price arbitrary authorizers,” it’s: do we actually need full programmability at this stage, or is there one generic authorizer FWSS can offer that unblocks the need?

so the requirements I’m reading 
- path-scoped permissions: is this a must?onchain?
- fast/ephemeral delegation, revocation: sounds reasonable  and the  default ACL should be meant to cover.

can we build that one authorized for now, benchmark the gas and define the ops fee; skip whitelist/per authorizer pricing; revisiting the programmability later if a 2nd use case shows up with genuinely incompatible auth needs? 

Feedback welcomed!

service_contracts/src/interfaces/IDataSetAuthorizer.sol
```diff
@@ -0,0 +1,28 @@
...
+    /// @param dataSetId The data set being operated on.
+    /// @param payer The data set's payer (the on-chain owner of the rails).
+    /// @param operation The operation type hash (e.g. SignatureVerificationLib.ADD_PIECES_TYPEHASH).
+    /// @param digest The EIP-712 digest signed for the operation.
+    /// @param signature The raw signature over `digest`; the authorizer recovers the
+    ///        signer itself, on whatever curve it supports.
+    /// @param operationData The ABI-encoded raw operation payload forwarded by FWSS, carrying all
+    ///        operation-specific data (e.g. piece metadata) so the authorizer can gate on its
+    ///        contents — for example enforcing metadata- or path-based ACLs. Empty for operations
+    ///        whose data is fully described by the other parameters (e.g. terminate).
```
Kubuxu, line 21: `raw operation payload forwarded` and `carrying all operation-specific data` could be at conflict here.
The raw operation payload suggests that the `operationData` should only be the information covered by the digest/signature but `carrying all operation-specific data` might be more than that.

Are there any such cases? Do we want to separate the data covered by the signature from any other supplemental data that FWSS (or maybe the uploader) might provide in the future?

Kubuxu:
Because of how generic the interface is, no concerns about limiting ourselves in the scope of authorisation schemes.  Although it also means that the authorizer becomes tied to the FWSS version. If the typed hash of the operation changes, it will require redeployments or upgrades to the authorizer.
Is that the tradeoff we want to make?

Admitting arbitrary authorizers exposes us to the griefing attack discussed above, and if we don't resolve that issue, providers will do so by whitelisting which authorizers they are willing to utilise during upload. I'm not saying that a whitelist is a good solution, but it is the easiest one to reach for.

Assuming some kind of whitelisting, can we keep the authorizers under clients' control while allowing for whitelisting based on code and upgrades?

Even if we were to deploy a single more constrained authorizer, I would suggest keeping the rough interface shape as long as we agree to make the first trade-off.

wjmelements:
Following discussion with @jennijuju, we will start with a gas cap on that call, in addition to the whitelist, in order to reduce our operational burden. The reason is that we will need to quantify the possible gas use anyway, as part of the approval process. Another benefit might be that the constraint will push developers to use less gas. This doesn't solve the pricing problem but it does prevent operational gas usage from becoming too excessive.

A complexity with this cap is that storage gas costs increase over time due to HAMT traversal, so we would still need to ensure these plugins are well below the limit. If they begin just under it, storage growth can push them over, and then transactions would start reverting. This rejection behavior is identical to Curio rejecting the requests due to high gas cost, which is the behavior I want anyway.

I still think we have to worry about network congestion and gas price in Curio. If SPs can't reject or delay messages during periods of high gas price, then they can be drained to DoS the network relatively cheaply.

I recommend 100 million gas for the limit. A single addPiece uses about 500 million gas right now. That should be more than enough headroom.

We also still need to agree on [where the whitelist will live](https://github.com/FilOzone/filecoin-services/pull/536#issuecomment-4908305101).

> I don't have a strong preference.

jennijuju:
> Assuming some kind of whitelisting, can we keep the authorizers under clients' control while allowing for whitelisting based on code and upgrades?



ah @Kubuxu  — do you mean we only care about the gas cap; but making sure the authorizer meets the protocol constraint is then integrators’s concern?

Kubuxu:
I mean the following:
1. I want us/others to be able to upgrade contracts that are fulfilling `IDataSetAuthorizer` as the method fingerprints can change.
2. But clients probably don't want the contracts to be upgraded from under them. They might want to vet and then upgrade. (product hypothesis).
3. If we implement whitelisting, we need to identify what code is being run for the authorisation contract in the face of upgrades and client-based ownership.

BigLep:
For visibility, a public design document has been drafted:
* Link to editable version: https://app.notion.com/p/filecoindev/202607-Optional-dataset-level-programmable-Authorizers-397dc41950c1809295ddc3d6ff6669b0
* Link to public read only version: https://app.notion.com/p/filecoindev/202607-Optional-dataset-level-programmable-Authorizers-397dc41950c1809295ddc3d6ff6669b0?source=copy_link

jennijuju:
I understand the use case better now and understand where the programmable requirement is coming from; and I can also see how the design here is extensible, in a good way. 

adding a bit more on my takeaway after discussing w/ Will on the gas cap;  I'm now comfortable with the cap-only approach for the time being — i.e. we don't need to add an authorizer ops-fee immediately, **as long as we size the gas cap so the uncompensated cost stays negligible against existing ops fees.** 

`isAuthorized` is called at each `Addpiece` and there already exists a  add-piece fee (0.0005 + 0.0003×N), so any gas isAuthorized adds comes straight out of the SP's fee from addpiece. Assuming my source is right lolol from 30-day data, a typical addPieces call burns ~3.45B gas at ~22 pieces/call, and a single-piece call is ~384M gas. So William's proposed 100M cap's impact is 

 | batch size | 100M cap = SP fee reduction |
  | --- | --- |
  | 1 piece (worst case) | ~26% |
  | 22 pieces (current avg) | ~3% |
  | 60 pieces (full batch) | ~1% |


At batch sizes we actually see, 100M is ~3%, and at a full batch ~1% — negligible, agreed. My unease is on the worst case: a client doing single-piece adds gives up ~26% of the SP's fee, and i prefer size to the worst case, not the average. I'd like the worst case constrained to ≤5% if we can, which points to a cap around ~20M gas (that's ~0.6% on an average batch).

However, the blocker on picking the number: the first authorizer I actually want to make sure we fully unblock — and the common case I'd like supported first — is P256 / passkey verification, and we don't yet know w
hat P256 costs on FEVM. (Im not sure if this is also Refound's priority, if not we also needs to unblock them ideally).  FIP-0113 adds the P256VERIFY precompile but deliberately assigns no gas number (FEVM meters by Wasm execution) and explicitly lists benchmarking it as an open TODO. Until someone benchmarks a representative P256VERIFY on calibnet, we can't say whether a passkey authorizer lives under a 5%-ceiling cap.

So two ways forward to unblock this for now:

1. Land conservative now, raise later. Ship with ~20M (≤5% worst case), and bump it via the normal upgrade path once we have the P256 benchmark and know what the real authorizers need. This is my preference — it keeps SP downside bounded while we learn, and loosening a cap later is low-risk.
2. Loosen now, accept more downside. Cap at 50M and accept a ~13% worst-case single-piece haircut. Simpler, gives P256 more headroom immediately, but I'm less comfortable committing SPs to that before we know we need it.

 Can we get that from the FIP-0113 implementers, or measure it on calibnet? That single figure tells us whether the passkey case fits a conservative cap or genuinely needs the higher-cap-plus-surcharge tier .

To be clear on the ops-fee question: I'm fine deferring it. The cap alone could be sufficient for now precisely because at these sizes the uncompensated gas is a rounding error on batched calls. We still need to revisit a surcharge if/when an authorizer genuinely requires a cap high enough to move SP margin.

wjmelements (approved):

essamhassan:
As discussed moved this the authorizer from external view to mutating call to leverage caching and reduce cost, this also unlock a number of interesting authorizer use-cases for other builders.

service_contracts/src/lib/SignatureVerificationLib.sol
```diff
@@ -309,13 +317,30 @@ library SignatureVerificationLib {
         uint256 dataSetId,
         address authorizer,
         bytes calldata operationData
-    ) public view returns (address) {
+    ) public returns (address) {
+        require(!_authorizerLocked(), Errors.AuthorizerReentrancy());
+        _setAuthorizerLock(true);
```
wjmelements: use `modifier` for before/after patterns like this

essamhassan: good call! addressed

wjmelements:
Merge `origin/main`, which bumps `forge` to `1.7.1`. Locally, do:
```sh
foundryup --update
foundryup --install v1.7.1
```

Resolve merge conflicts and rerun `make gen`, `forge fmt`, `forge lint` with `v1.7.1`, since those have changed.

JAG-UK (approved):
I have reviewed and fully exercised this code with my own devnet and home-grown Authorizer and client app. It's compatible with direct sig delegation and Passkeys and is an elegant solution to on-chain auth for FWSS operations.

It does have a couple of high-impact implications on the core project:
* P256 signatures add around 100M gas(!) This is within expected bounds but it's still a lot. On its own the mechanism doesn't force the gas up but as soon as you use it it will, and it's up to the CLIENT to make and register the verifier while the SP pays the gas. There's a 150M max griefing limit in here, which is good, but we need to assume that it will always hit that limit and double-check we're OK with that cost.
* extradata size limits need to increase a little bit - 512 to 1024 bytes for each operation type.

service_contracts/src/lib/SignatureVerificationLib.sol
```diff
@@ -291,4 +292,70 @@ library SignatureVerificationLib {
...
+    ///      blocked. The pre-`_;` check-and-set is hoisted into a function so the modifier body stays a plain
+    ///      sequence of calls (forge-lint `unwrapped-modifier-logic`).
+    modifier nonReentrantAuthorizer() {
+        _lockAuthorizer();
+        _;
+        _setAuthorizerLock(false);
+    }
+
+    /// @notice Delegates the authorization decision for an operation to the data set's authorizer.
+    /// @dev Called only when an authorizer is attached: it is the sole gate. FWSS forwards the raw signature
```
JAG-UK: > it is the sole gate

I wonder if there isn't a benefit to having a fast 'caller == payer' check before any of this? In reality the payer can always do anything, but as this stands as soon as the Authorizer is set then you have to add code or processes to verify the payer when we already have a good secp256k1 fast verification path.

rvagg:
> extradata size limits need to increase a little bit - 512 to 1024 bytes for each operation type

This is slightly outdated now because we implemented the limitations that were added at the bottom, and we're going to be working further on unblocking limits in https://github.com/FilOzone/pdp/pull/292 and a follow-up FWSS eventually, but it's still relevant: https://app.notion.com/p/filecoindev/addPieces-Batch-Limits-and-Proposed-Changes-36ddc41950c180d39f45f0da8982090c

There should be numbers in there that would help you figure out what size a maxxed out batch addPieces is and therefore what this new overhead is going to add. The main limit we'll hit immediately is the message size limit. IIRC the hard limit is currently 41 and we coded in 40 as a nicer round number into Curio and the SDK. Perhaps the new overhead will push this down even further. We don't necessarily have to restrict the batch size further to accommodate this because it's still theoretical and people don't typically do heavy metadata in these so we have headroom, but what it does mean is new failure conditions to explain to users: _why can't I batch when you said I could do 40 at a time?_ Sadly it's an obscure edge case that is annoying to have to even explain, but one day someone will hit it.

service_contracts/src/lib/SignatureVerificationLib.sol
wjmelements: The payer isn't allowed to call these operations; they require consent of the SP.

Kubuxu:
@rvagg, as part of https://github.com/FilOzone/filecoin-services/issues/575, I suggest we implement smart batch limits so we can detect early which batch size will work for each use case, maximising batching.

Kubuxu (approved):

Kubuxu:
The contract size check failing is "bogus", as it is a harness wrapper of FWSS that breaches it, but this PR is getting quite close to capping it out.

https://gist.github.com/Kubuxu/da8898bbeb7babc05688c500f66400a7
Increase by 1695 from 22863 to 24558, which is 18 bytes below the limit of 24576, not leaving us room for even bug fixes.
In a call, we discussed that we should reserve at least 1KiB for bug fixes.

Kubuxu:
I moved the entire dispatch into the signature library, which significantly reduces the size increase (1700 -> 134).
Gas usage takes a slight hit, but it is manageable (0.5%).

wjmelements (approved):

JAG-UK:
> I moved the entire dispatch into the signature library, which significantly reduces the size increase (1700 -> 134). Gas usage takes a slight hit, but it is manageable (0.5%).

Great catch thanks @Kubuxu

PR #540, opened 2026-06-29
feat: ExampleSponsoredDataSet

### LINKED ISSUE #531: implement contract factory for "sponsor a dataset" functionality
Full design (Notion, internal): https://app.notion.com/p/filecoindev/PRD-Design-Plan-Sponsor-Vault-Factory-Option-A-380dc41950c1819bb0a6f46000880a6c

## Problem

People want to pay for storage of data they don't own: keep a public dataset alive, sponsor a grantee's storage, let an app fund storage for its users. Filecoin Pay can't express this today. Balance is pooled per (token, account), so funding a wallet that backs many datasets gives a donor no guarantee their money serves the dataset they cared about.. dataset 999 can drain what you meant for dataset 789.

This keeps coming up from clients and ecosystem partners, and the answer each time is the same: make the payer a contract.

## What we're building

A `SponsorVaultFactory` that deploys cheap, immutable minimal-proxy vaults, one vault per dataset. The vault becomes the Filecoin Pay payer for exactly one dataset. Anyone can `deposit()` into it; nobody can withdraw mid-life. The earmark holds because the vault pays for nothing else.

A contract can't sign EIP-712, so the vault registers a steward EOA as scoped session keys (create-dataset, add-pieces, remove-pieces) to operate the dataset. The steward runs the dataset and never controls funds.

Funds leave a vault only three ways: settlement to that dataset's SP, `migrate()` to a successor vault for the same dataset, or `refund()` to the treasury as a last resort. No discretionary withdrawal, including by the deployer. That no-withdraw guarantee is the whole point.. `withdraw` on Filecoin Pay releases balance minus lockup, and Warm Storage only locks ~30 days, so any owner-controlled withdraw could drain ~23 of 24 months of a prefund.

## v1 scope

PDP-only, direct deposit (no per-donor ledger or refunds), one dataset per vault, immutable (no upgradeability). The explorer/donation UI resolves `datasetId -> vault` from the factory's `VaultDeployed` events, not an on-chain registry.

Not in scope: CDN / sponsored retrieval, rail-scoped balances in Filecoin Pay, retrofitting existing datasets, multi-dataset vaults.

## Build plan

- [ ] Spike (1-2 days) on calibnet: vault funded first, SP-submitted creation with steward session-key signature, bind, add pieces, terminate, then migrate or refund. Everything gates on this.
- [ ] `SponsorVaultImpl` + `SponsorVaultFactory` with invariant tests (no withdraw path, migrate only to a validated same-dataset successor, no key moves funds to an arbitrary address).
- [ ] Audit scoped to the no-withdraw guarantee, migrate/refund gating, one-dataset invariant.

wjmelements:
Closes #531
[Design Doc](https://app.notion.com/p/filecoindev/PRD-Design-Plan-Sponsor-Vault-Factory-Option-A-380dc41950c1819bb0a6f46000880a6c)
Reviewer @ZenGround0
CC @essamhassan @SgtPooki @rvagg @kubuxu
#### Context
We want to start iterating on this use case before replication+repair and before FPv2 since it seems several customers want this.
#### Overview
This implements a smart contract payer for a singular data set.
The payer is funded by deposits to its FilecoinPay account or by FilecoinPay Rails.
##### Motivation
One goal is to minimize custody: donated funds can only pay for the data set.
Another goal is to allow finalization: funds paying for an immutable data set can only pay for that data set.
##### Participants
Anyone can donate to sponsor the data set.
There is a curator, who can add or remove pieces from the data set.
The curator can abdicate, which finalizes the data set.
There is also a fallback beneficary, who receives the funds in the event the data set is lost.
##### Migration
Migration is only allowed to sibling SponsoredDataSets and requires proof of data possession in the recipient.
For finalized data sets we require total data set equality.
#### Changes
* add SponsoredDataSet factory and children
* add tests


service_contracts/src/lib/FilecoinWarmStorageServiceStateLibrary.sol
```diff
@@ -97,6 +97,19 @@ library FilecoinWarmStorageServiceStateLibrary {
...
+            return false;
+        }
+        uint256 maxPeriod = getMaxProvingPeriod(service);
+        uint256 currentPeriod = (block.number - activation - 1) / maxPeriod;
+        if (currentPeriod >= 1 && provenPeriods(service, dataSetId, currentPeriod - 1)) {
+            return true;
+        }
+        return provenThisPeriod(service, dataSetId);
+    }
+
```
JAG-UK, line 110: I'm exposing my ignorance of the FWSS architecture here but do we really need to have a perfect (other than view mode) clone+hack of the function in the internal contract here? Can't it be shared? There's no state modification or anything in here and then one change in the future correctly fixes every caller.

service_contracts/src/lib/FilecoinWarmStorageServiceStateLibrary.sol
```diff
@@ -124,6 +137,19 @@ library FilecoinWarmStorageServiceStateLibrary {
...
+        public
+        view
+        returns (address payer, uint256 pdpRailId)
+    {
+        bytes32 slot = keccak256(abi.encode(dataSetId, StorageLayout.DATA_SET_INFO_SLOT));
+        bytes32[] memory info4 = service.extsloadStruct(slot, 4);
+        pdpRailId = uint256(info4[0]);
+        payer = address(uint160(uint256(info4[3])));
+    }
+
```
JAG-UK, line 150: I'm exposing my ignorance of the FWSS architecture here but do we really need to have a perfect (other than view mode) clone+hack of the function in the internal contract here? Can't it be shared? There's no state modification or anything in here and then one change in the future correctly fixes every caller.

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,355 @@
...
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SignatureVerificationLib} from "./lib/SignatureVerificationLib.sol";
+import {FilecoinWarmStorageServiceStateInternalLibrary} from "./lib/FilecoinWarmStorageServiceStateInternalLibrary.sol";
+import {PDPVerifier} from "@pdp/PDPVerifier.sol";
+import {LibRLP} from "solady/utils/LibRLP.sol";
+import {LibBytes} from "solady/utils/LibBytes.sol";
+import {LibClone} from "solady/utils/LibClone.sol";
+import {FVMPay} from "@fvm-solidity/FVMPay.sol";
+
+function hashTypedDataV4(FilecoinWarmStorageService fwss, bytes32 structHash) view returns (bytes32) {
```
wjmelements, line 16: This function will also let us simplify some of our SignatureVerificationLib

service_contracts/src/lib/FilecoinWarmStorageServiceStateLibrary.sol
wjmelements: The duplications are generated code. There are three copies for three different types of consumer:

* public library
* internal library
* public view contract

The SponsoredDataSet only uses the internal library. If we needed to cut codesize we could use a public library. The view contract is used mainly for `eth_call`.

service_contracts/src/lib/FilecoinWarmStorageServiceStateLibrary.sol
wjmelements: https://github.com/FilOzone/filecoin-services/pull/540#discussion_r3493555643

JAG-UK:
Thanks @wjmelements I like this a lot: the principle that a smart contract acts as payer for a dataset is certainly cleaner than other ideas like handing over active rails and things like that.

The one big missing piece for most of my use cases is that we are concerned with very large data sets - like 10s of TiBs at least - which means P2PP DDO PoRep deals for the foreseeable future. This PR is very tightly coupled with FWSS and therefore not a good fit for this use case.

Conceptually it's very close though. In P2PP deals are funded with Filecoin Pay rails and the NFRs of not wanting to deal with custody, maintaining tight control of migration etc all apply pretty much exactly as implemented here. We are planning to integrate P2PP PoRep with FoC very soon as an archival offering ("Filecoin Cold Storage Service"). Do you think it's reasonable to try to make this PR a little more flexible to talk about deals and CIDs in multiple dialects? Or would you rather deploy this just for FWSS PDP deals and we take it as inspiration to deploy a separate parallel system for FCSS?

JAG-UK:
One other system-wide thought: will there be a problem making a smart contract the payer in terms of identifying who 'owns' a dataset for the purposes of signing vouchers for retrieval or applying UCAN access control? Are we comfortable that the designs for those parts of the puzzle will play nice with this model? 

My immediate use cases for sponsored large datasets are public goods so none of this applies, but things can grow and I don't want us to miss an obvious face rake if we can avoid it.

wjmelements:
> One other system-wide thought: will there be a problem making a smart contract the payer in terms of identifying who 'owns' a dataset for the purposes of signing vouchers for retrieval or applying UCAN access control? Are we comfortable that the designs for those parts of the puzzle will play nice with this model? 
> 
> 
> 
> My immediate use cases for sponsored large datasets are public goods so none of this applies, but things can grow and I don't want us to miss an obvious face rake if we can avoid it.

Our current plan for UCAN support is https://github.com/FilOzone/filecoin-services/pull/536 which also allows secp256r1 and other custom auth.

jennijuju:
Want to set expectations on this PR from the product side, because the scope here is narrower than the request that started it — and that narrowing happened without product in the loop.

The customer ask was "can our users sponsor a rail," meaning fund both storage and egress on a given dataset of FWSS. The general idea we discussed in NYC was: make the payer of a rail a contract and any service on that rail can be sponsored. This PR doesn't deliver that. It's a stopgap scoped to FWSS PDP rails only. That's fine as a first step, but let's be explicit about what it is and isn't so nobody downstream over-reads it (correct me if im wrong):
- **It is not a reusable sponsorship primitive.** : It's FWSS/PDP-coupled and immutable; other services (egress, FCSS/P2PP — see JAG-UK's comment) each need their own contract.
- It does not mean "sponsored service" is now a FOC capability.
  -  : delivering the actual "sponsor a rail" request — storage and egress, reusable across services — needs FilecoinPay v2 with depositToRail (rail-scoped balances). Is that the next step, and is the egress rail dependency in its scope? I don't want us to ship this, call "sponsorship" done, and quietly leave the real ask to a v2 that isn't sequenced.
-  It's plumbing for integrators who onboard datasets  and want easier donations toward storage — **not a self-serve feature for end storage users.** Someone still deploys, binds, and operates a vault per dataset.
- Most important, it does not serve the egress half of the customer ask.

The design doc suggests the CDN rail is out of scope,  however, i read the PR correctly,  a CDN-enabled dataset would have egress bleed the same deposits with no safe guard still.  How is that managed atm?

Net: I'm fine shipping this as an explicitly-scoped storage-only  sponsorship for FOC. I'm less fine if it's positioned as "Sponsor/Sponsor service is supported on FOC "  — because the request was sponsored service, and that's still ahead of us.

wjmelements:
> a CDN-enabled dataset would have egress bleed the same deposits with no safe guard still. How is that managed atm?

That's true. Data sets in FWSS are tightly bound to their FilBeam add-on. The same client must pay for both; there isn't a way to pay for one but not the other. The way it is handled is that you can turn it off in FWSS and pay for it out of band. Whether it is turned on or off, this setting is persisted to successor SponsoredDataSets, so if you donated to pay for FilBeam, you will keep paying for FilBeam, and if you did not you will not.

> it does not serve the egress half of the customer ask.

Does this mean:

> The customer ask was "can our users sponsor a rail," meaning fund both storage and egress on a given dataset of FWSS

This feature delivers funding both storage and egress on a FWSS dataset. What it does not do is allow "deposit to rail" since rails don't have accounts in FilecoinPay (even though that would have been a better way to implement lockups, particularly with [ERC-6909](https://github.com/FilOzone/filecoin-pay/issues/218)).

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

Terrajlz:
I'd like to work on this.

Recent work:
- Nobay Protocol: 21/21 Hardhat tests passing, 15-finding security audit, full OZ v5 migration
- DeXe Protocol: identified High-severity tx.origin vulnerability in PoolFactory (SWC-115)
- Audited contracts across Aave V3, GMX, Uniswap V3, MakerDAO, Lido, ENS, Ethena

I can deliver clean, tested code with security audit. Turnaround: 24 hours. Ready to start immediately.

wjmelements:
Per discussion with @rvagg today we're renaming this to ExampleSponsoredDataSet to signal that this is a template for developers to use as a starting point.

rvagg:
This is good by me, it just needs a doc somewhere - either in the readme or alternatively an "examples" directory with its own readme and this file in it.

Unfortunate we have a new dependency but it's only for tests at least.

parkan:
sorry for all the delays in looking at this, was AFK at dwebcamp (still at camp for breakdown but feeling bad about missing the review request so sneaking away to review)

EDIT: hmm I see some issues here, will post analysis shortly

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,365 @@
...
+        string[] calldata metadataKeys,
+        string[] calldata metadataValues,
+        address curator,
+        address beneficiary
+    ) external returns (ExampleSponsoredDataSet) {
+        bytes32 createDataSetStructHash =
+            SignatureVerificationLib.createDataSetStructHash(0, payee, metadataKeys, metadataValues);
+        bytes32 createDataSetHash = hashTypedDataV4(WARM_STORAGE_SERVICE, createDataSetStructHash);
+        // Sign CreateDataSet using Nick's Method
+        address createDataSetSigner = ecrecover(createDataSetHash, 0, 0, 0);
```
parkan: this recovers address(0) due to invalid parameters so in `initialize` any registered SP can create datasets billed to the contract and lock/drain funds, needs a real signer -- I don't think this is intentional?

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        emit MigrationInvalid(migrationId, pieceIndex);
+        uint256 half = MIGRATION_DEPOSIT / 2;
+        require(FVMPay.pay(msg.sender, half), PayFailed());
+        require(FVMPay.burn(half), BurnFailed());
+    }
+
+    /// @notice Completes a pending migration after the challenge period has elapsed.
+    /// @dev Transfers available funds to the successor and refunds the deposit to the original depositor.
+    function completeMigration(uint256 migrationId) external {
```
parkan, line 293: you need to `_checkMigrationConditions` here again otherwise you can propose a successor, immediately delete and collect the bond without challenges

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,365 @@
...
+        SESSION_KEY_REGISTRY.login(curator, type(uint256).max, curatorPerms, ORIGIN);
+        PAYMENTS.setOperatorApproval(
+            TOKEN, address(WARM_STORAGE_SERVICE), true, type(uint256).max, type(uint256).max, type(uint256).max
+        );
+
+        CURATOR = curator;
+        BENEFICIARY = beneficiary;
+    }
+
+    function bind(uint256 _dataSetId) external {
```
parkan, line 169: this is external so it needs access controls and a sanity check on _dataSetId

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,365 @@
...
+    function _resolveSuccessor(ExampleSponsoredDataSetFactory factory, uint64 thisNonce, uint64 successorNonce)
+        internal
+        view
+        returns (ExampleSponsoredDataSet)
+    {
+        require(LibRLP.computeAddress(address(factory), thisNonce) == address(this), FactoryMismatch());
+        return ExampleSponsoredDataSet(LibRLP.computeAddress(address(factory), successorNonce));
+    }
+
+    function _checkMigrationConditions(ExampleSponsoredDataSet successor)
```
parkan, line 318: this should probably check `successor.BENEFICIARY` like in `migrateUnfinalized`

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,365 @@
...
+        );
+        require(
+            PDP_VERIFIER.getNextPieceId(dataSetId) == PDP_VERIFIER.getNextPieceId(successorDataSetId), PieceMismatch()
+        );
+    }
+
+    function _transferFunds(ExampleSponsoredDataSetFactory factory, ExampleSponsoredDataSet successor) internal {
+        IERC20 token = factory.TOKEN();
+        (uint256 funds, uint256 lockupCurrent,,) = PAYMENTS.accounts(token, address(this));
+        uint256 available = funds - lockupCurrent;
```
parkan: this should probably settle before doing the math? reverts if we overcounted in the snapshot

service_contracts/src/ExampleSponsoredDataSet.sol
```diff
@@ -0,0 +1,365 @@
...
+
+        successorDataSetId = successor.dataSetId();
+        require(successorDataSetId != 0, SuccessorNotBound());
+
+        require(
+            PDP_VERIFIER.getDataSetLastProvenEpoch(successorDataSetId) > successor.finalizedEpoch(),
+            SuccessorNotProven()
+        );
+        require(
+            PDP_VERIFIER.getNextPieceId(dataSetId) == PDP_VERIFIER.getNextPieceId(successorDataSetId), PieceMismatch()
```
parkan, line 348: I think this might break migrations if pieces were ever deleted -- repacking the surviving pieces after a deletion results in non-matching NextPieceId (deleted ids are not reused) and this reverts

service_contracts/src/ExampleSponsoredDataSet.sol
parkan: *collect unlocked funds, sorry

parkan (changes requested):
ok there's a few issues here that combine into a bad path:
- pick a set
- create a "matching" successor set with matching leaf count but without real data
- propose migration and delete successor in a single block/bundle, no one can challenge because this is basically atomic
- challenges revert because there's no dataset to compare with
- completeMIgration happily runs
- we take the unlocked funds from the source and release

as far as I can tell this is only guarded by registered status

other issues/more details in comments

service_contracts/src/ExampleSponsoredDataSet.sol
parkan: fixing this will close the above attack path but we should probably compare piece commitments *at propose time* and allow challenges against that, or see my comment in main thread below for a proper fix

parkan:
re:
> For finalized data sets we require total data set equality.

this is only actually checked here: 
https://github.com/FilOzone/filecoin-services/pull/540/changes#diff-e4c4ab4e9a41f0966f2475cad1096f7894ace1f3d02566fbec4691fdbe1ffb10R359

the propose flow does not -- this is the "optimistic" path that avoids enumerating pieces but as it stands now is exploitable

(everything below is speculative/needs more work, fixing the above closes the immediate gaps for now)

the real solution would be for PDPVerifier to expose a merkle root per set with tombstones filtered (getActivePieces already does this so just need an amortized set-root update, this might even exist somewhere but I can't find it) -- this way all the migration dances would reduce to O(1) fully deterministic and provable comparisons

EDIT: actually the real/simple fix is even simpler: since only finalized datasets may be migrated then we only need to compute the root once, at finalize time

my guess is that this was considered and rejected for some concrete reason, but the merkle root computation should already be optimized and doing so once upon finalize should not be expensive relative to running all the fraud proof machinery -- let me know if I'm overlooking something here

EDIT2: ok the other other observation is that beyond tombstone count making sets unequal even if alive pieces are exactly the same and in the same order (by the nextId metric) the thing that we *actually* want is an _unordered_ set equality -- the order of addition is only meaningful incidentally, it does not reflect contributor intent in any case I can think of, so the construction that we want is not a strict merkle root but an order-independent multiset like MuHash in BTC

service_contracts/src/ExampleSponsoredDataSet.sol
parkan: (ok it might be intentional and I am not understanding the intended mechanics? we do check for address(0) in a bunch of other places)

parkan: ok if you're actually trying to use Nick's Method you need to pick a valid random (r,s) -- zeroes are not valid

you need to guard/revert on recovered address(0) and try to pick valid inputs

wjmelements:
> PDPVerifier to expose a merkle root

I think it's too late to add that kind of accounting. I also considered a trie for the proof system. Data "sets" can have duplicates though, and there's also piece metadata which would ideally also match.

service_contracts/src/ExampleSponsoredDataSet.sol
wjmelements: > valid

yes

> random

no

> need to guard/revert on recovered address(0)

I don't think so. `r = 1` signatures are always valid.

4409d424862f89a113f90af474fd392f56923897

parkan:
> > PDPVerifier to expose a merkle root
> 
> I think it's too late to add that kind of accounting. I also considered a trie for the proof system. Data "sets" can have duplicates though, and there's also piece metadata which would ideally also match.

yeah, understandable -- thus it's under the speculative section (and yes, merkle is not really the right aggregation here, just the first tool on the shelf in PL-land)

service_contracts/src/ExampleSponsoredDataSet.sol
wjmelements: I went with a successor liveness check 09f05ba02fd1f745920c652a84b58ce3c9ff2538

service_contracts/src/ExampleSponsoredDataSet.sol
wjmelements: This method does not require any access control or sanity checks. It is impossible to bind any data set except the singular authorized one.

Claude explains:

```
  bind(uint256) is not unsafe — despite having no access control (external, callable by anyone) and no dataSetId == 0 guard against being called twice.

  Only one data set can ever exist with payer == address(this). In FilecoinWarmStorageService.dataSetCreated (src/FilecoinWarmStorageService.sol:547-551), a new data set is only
  registered for (payer, clientDataSetId) if clientNonces[payer][clientDataSetId] == 0; it's then set permanently (and explicitly not cleared on deletion — see the comment at line
  680, "keep clientNonces... to prevent replay"). The ExampleSponsoredDataSetFactory flow always uses clientDataSetId = 0 (src/ExampleSponsoredDataSet.sol:60), so that slot for a
  given clone address is a one-shot: once consumed, no second data set — from the factory flow or from any other attacker-crafted PDPVerifier.createDataSet call with a forged payer —
  can ever be registered with payer == address(this).

  The payer field is caller-supplied, not derived from msg.sender (src/FilecoinWarmStorageService.sol:535-559 decodes it straight from extraData), so in principle anyone could try to
  create a data set claiming payer = <some ExampleSponsoredDataSet address>. But it's gated by verifyCreateDataSetSignature, which requires a signature recovering to an address the
  SessionKeyRegistry has authorized for that payer with CREATE_DATA_SET_TYPEHASH. The only such authorized signer is the Nick's-Method pseudo-address ecrecover(hash, 27, 1, 1)
  computed from one specific hash (clientDataSetId=0, payee, metadata) at initialize() time — since ECDSA recovery is a one-to-one function of the digest for fixed (v,r,s)=(27,1,1),
  that same authorized address cannot be reproduced for any other digest, so no attacker can get a different createDataSet call to pass the signature check for this payer.

  So concretely:
  - Can a second data set ever be created and bound? No — the nonce guard makes the one legitimate data set (clientDataSetId=0) the only one that will ever have payer ==
  address(this), permanently, even after deletion.
  - Can bind() be tricked into binding a different data set than intended? No — any _dataSetId other than the legitimate one will have payer != address(this) and bind() reverts. The
  legitimate one can only be "rebound" to itself (a no-op).

  The lack of an explicit dataSetId == 0 check or curator-only gate on bind() is a style looseness, not an exploitable vulnerability — the real invariant is enforced upstream by
  FWSS's per-(payer, clientDataSetId) nonce, which the contract correctly relies on.
```

service_contracts/src/ExampleSponsoredDataSet.sol
wjmelements: 45fcdae2a5ab2e96d1dec2b129a0a44e79c5b90a

service_contracts/src/ExampleSponsoredDataSet.sol
wjmelements: 0c23f0412a039e8f2c8f7b9b85353bbf8ee1b7c8

SgtPooki:
@rvagg @Kubuxu @ZenGround0 since this is just an example, can we get an approval here so we can merge in?

rvagg (approved):

Kubuxu:
Hey, this PR resulted in a bad merge, so it was reverted in #599.
Please reopen it, and we will get it merged after the release.
Also, there is a TODO left in the StateView

PR #543, opened 2026-07-07
fix(fwss): clear scheduled metadata removals on dataset deletion

rvagg:


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -711,6 +711,8 @@ contract FilecoinWarmStorageService is
         }
         delete dataSetMetadataKeys[dataSetId];
 
+        processScheduledPieceMetadataRemovals(dataSetId);
```
wjmelements, line 714: I suppose it's good to clear these out. Too bad we can't also clear out the metadata for the pieces that aren't scheduled for deletion.

wjmelements (approved):

PR #545, opened 2026-07-09
fix: PDP rail settlement when endEpoch equals activationEpoch

### LINKED ISSUE #544: FWSS settlement can become impossible when rail endEpoch == activationEpoch
FilecoinWarmStorageService.validatePayment has an inconsistent boundary condition for settlement ranges that end exactly at the dataset activation epoch.

```solidity
_findProvenEpochs allows:
toEpoch >= activationEpoch
```

but 

```solidity
_provingPeriodForEpoch treats the same epoch as invalid:
epoch <= activationEpoch
```

and 

```solidity
returns type(uint256).max
```

When a terminated PDP rail has endEpoch == activationEpoch, settleRail(railId, endEpoch) passes the outer range check but then sends the sentinel into the proving-period loop. This causes gas estimation/execution to run out of gas instead of cheaply settling or rejecting the range.
For this rail state there is no usable validator settlement target: toEpoch == activationEpoch hits the gas-bomb path, while any lower toEpoch reverts with InvalidEpochRange.

LexLuthr:
fixes https://github.com/FilOzone/filecoin-services/issues/544

## Summary

Fixes an FWSS payment validator boundary bug where a terminated PDP rail could become impossible to settle when `endEpoch == activationEpoch`.

`_findProvenEpochs` allowed `toEpoch == activationEpoch`, but `_provingPeriodForEpoch` treats `activationEpoch` as outside any proving period. That caused the validator to derive `endingPeriod = type(uint256).max` and enter an effectively unbounded proving-period loop.

The fix handles the exact activation boundary before proving-period calculation, returning zero payment and `settleUpto = activationEpoch`.

## Behavior

- `toEpoch < activationEpoch` still reverts `InvalidEpochRange`
- `toEpoch == activationEpoch` now settles to the activation boundary with zero payment
- `toEpoch > activationEpoch` keeps the existing proving-period validation logic

## Tests

- Added direct validator coverage for `(activationEpoch - 1, activationEpoch]`
- Added terminated-rail regression coverage for `endEpoch == activationEpoch`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 
 # git worktrees
 .trees/
+
+.idea
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -1553,6 +1553,9 @@ contract FilecoinWarmStorageService is
             fromEpoch = activationEpoch;
         }
         settleUpTo = fromEpoch;
+        if (toEpoch == activationEpoch) {
+            return (0, settleUpTo);
+        }
         uint256 startingPeriod = _provingPeriodForEpoch(activationEpoch, fromEpoch + 1, maxProvingPeriod);
         uint256 endingPeriod = _provingPeriodForEpoch(activationEpoch, toEpoch, maxProvingPeriod);
         uint256 deadline = _calcPeriodDeadline(activationEpoch, startingPeriod);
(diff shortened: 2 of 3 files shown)


wjmelements (approved):

rvagg:
Follow-up https://github.com/FilOzone/filecoin-services/pull/552

Issue #541, opened 2026-07-01
Enforce a minimum delay for FWSS and PDPVerifier contract upgrades


BigLep:
## Summary

Add a contract-enforced minimum delay between announcing and executing upgrades to the FWSS and PDPVerifier contracts.

The FOC WG considered doing this as part of #505 / #533 and decided on 2026-06-30 **not to prioritize contract-enforced delay now**. For the immediate work, we will keep #505 focused on contract/operator ergonomics. This issue preserves the security-hardening work as an explicit backlog item for future prioritization.

Decision record and trade-off analysis: [202606 Contract upgrade delay decision: FWSS and PDPVerifier](https://app.notion.com/p/filecoindev/202606-Contract-upgrade-delay-decision-FWSS-and-PDPVerifier-38bdc41950c18080be5bfe1179b53a81)

## Why

A contract-enforced delay provides a verifiable security guarantee that process alone cannot provide. It gives operators and integrators time to inspect an announced implementation and react to deployment mistakes, malicious upgrades, or a compromised/social-engineered multisig before the new implementation can execute.

We are deferring it because current customer demand is limited, FOC still expects meaningful contract iteration, and the delay would constrain operational response. We should revisit that balance as adoption, funds at risk, and demand for trust minimization grow.

## Direction discussed

The leading future shape discussed was:

- Apply to both FWSS and PDPVerifier.
- Use a simple ramp that reaches approximately 48 hours after a period of stability, rather than the original two-week maximum proposed in #533.
- Keep the policy understandable and easy to communicate.
- Do not assume that rollback can always bypass the delay; rollback may be storage-unsafe.
- Preserve the ability to tighten the delay over time without needing to weaken an already-promised security guarantee.

The exact curve, grace period, and implementation remain to be decided when this is prioritized.

## Done criteria

- [ ] Confirm the current customer/integrator requirement and threat model.
- [ ] Decide the delay shape, ramp period, and maximum delay.
- [ ] Define behavior for emergency fixes and rollbacks.
- [ ] Implement the policy for both FWSS and PDPVerifier.
- [ ] Emit sufficient event data for operators and monitoring systems to determine the earliest execution epoch.
- [ ] Add contract tests, deployment/tooling support, and operator documentation.
- [ ] Coordinate with [FilOzone/tpm-utils#51](https://github.com/FilOzone/tpm-utils/issues/51) so upgrade announcements are surfaced in Slack. This monitoring improves transparency but does not itself provide a security guarantee.

[...]

PR #547, opened 2026-07-10
fix: non-reverting upgrade announcement

### LINKED ISSUE #505: Make announcePlannedUpgrade more forgiving for emergency rollouts
## Problem

During the [2026-05-28 Hotfix (FWSS v1.2.0 / PDPVerifier v3.4.0)](https://app.notion.com/p/filecoindev/2026-05-28-Hotfix-FWSS-v1-2-0-compatibility-with-PDPVerifier-v3-4-0-901dc41950c1820d99c88157214dec5d), the team was slowed down getting critical fixes deployed because `announcePlannedUpgrade` requires a two-step process (announce, then execute after the announced epoch) when using multisig wallets. In an emergency, this delay adds unnecessary friction — each step requires multisig coordination, and the epoch constraint makes it difficult to execute quickly.

The goal is to **enable quicker emergency rollouts** so that when a critical fix is needed, the upgrade mechanism doesn't become a bottleneck.

## Options discussed

Several approaches have been considered (see [Notion discussion](https://app.notion.com/p/filecoindev/announcePlannedUpgrade-is-more-forgiving-around-update-epoch-for-emergency-rollouts-371dc41950c180a29f3bcf47dfee8552)):

1. **Relax epoch constraints** — allow the upgrade epoch to be the current epoch or up to X epochs in the past, so the announce+execute flow is more forgiving of timing.
2. **Emergency mode parameter** — an explicit flag that makes it clear we are attempting an immediate upgrade, bypassing the normal delay.
3. **Replace `afterEpoch` with `minDelay`** — simplify coordination by specifying a minimum delay rather than a target epoch. The `minDelay` must exceed a `requiredDelay` that starts short immediately after deployment (facilitating emergency response) and gradually increases over time to a maximum.

Additional considerations:
- Recording the block/epoch in which each upgrade occurred (`mapping(version => epoch)`) for auditability and backwards compatibility.
- Whether to support reverts to a previous version (though this isn't always storage-safe).

## Next step

A quick design decision is needed on which approach to take before implementation begins. This affects the contract interface and upgrade coordination tooling, so the design should be agreed upon first.

After we agree on the approach, this will involve PRs to pdp and filecoin-services repos.

## Related

- Notion tracking: [announcePlannedUpgrade is more forgiving around update epoch for emergency rollouts](https://app.notion.com/p/filecoindev/announcePlannedUpgrade-is-more-forgiving-around-update-epoch-for-emergency-rollouts-371dc41950c180a29f3bcf47dfee8552)
- Automate Safe upgrade proposal: #436

wjmelements:
Supercedes #533 (no minimum delay)
Toward #505
Reviewer @ZenGround0
CC @rjan90
#### Changes
* do not revert if upgrade plan has low block number
* deprecate previous method specifying block number in favor of new method specifying delay
* test both methods using (bool) param

### DIFF
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -294,6 +294,7 @@ contract FilecoinWarmStorageService is
     // The address allowed to terminate CDN services
     address private filBeamControllerAddress;
 
+    // Pending upgrade announcement
     PlannedUpgrade private nextUpgrade;
 
     // Pricing rates (mutable for future adjustments)
@@ -405,11 +406,27 @@ contract FilecoinWarmStorageService is
         challengeWindowSize = _challengeWindowSize;
     }
 
-    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external onlyOwner {
-        require(plannedUpgrade.nextImplementation.code.length > 3000);
-        require(plannedUpgrade.afterEpoch > block.number);
-        nextUpgrade = plannedUpgrade;
-        emit UpgradeAnnounced(plannedUpgrade);
+    function announceUpgradePlan(address nextImplementation, uint96 delay) external {
+        if (delay == 0) {
+            delay = 1;
+        }
+        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delay);
+    }
+
+    /// @custom:deprecated Use announceUpgradePlan instead
+    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external {
+        uint96 minAfterEpoch = uint96(block.number + 1);
+        _announcePlannedUpgrade(
+            plannedUpgrade.nextImplementation,
+            plannedUpgrade.afterEpoch < minAfterEpoch ? minAfterEpoch : plannedUpgrade.afterEpoch
+        );
+    }
+
+    function _announcePlannedUpgrade(address nextImplementation, uint96 afterEpoch) internal onlyOwner {
+        require(nextImplementation.code.length > 3000);
+        nextUpgrade.nextImplementation = nextImplementation;
+        nextUpgrade.afterEpoch = afterEpoch;
+        emit UpgradeAnnounced(nextUpgrade);
     }
 
     function _authorizeUpgrade(address newImplementation) internal override onlyOwner {
--- a/service_contracts/src/FilecoinWarmStorageService.sol
+++ b/service_contracts/src/FilecoinWarmStorageService.sol
@@ -406,11 +406,11 @@ contract FilecoinWarmStorageService is
         challengeWindowSize = _challengeWindowSize;
     }
 
-    function announceUpgradePlan(address nextImplementation, uint96 delay) external {
-        if (delay == 0) {
-            delay = 1;
+    function announceUpgradePlan(address nextImplementation, uint96 delayEpochs) external {
+        if (delayEpochs == 0) {
+            delayEpochs = 1;
         }
-        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delay);
+        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delayEpochs);
     }
 
     /// @custom:deprecated Use announceUpgradePlan instead
--- a/service_contracts/src/ServiceProviderRegistry.sol
+++ b/service_contracts/src/ServiceProviderRegistry.sol
@@ -772,12 +772,12 @@ contract ServiceProviderRegistry is
     /// @notice Announce a planned upgrade
     /// @dev Can only be called by the contract owner
     /// @param nextImplementation Address of the new implementation contract
-    /// @param delay Number of epochs from now before the upgrade may occur
-    function announceUpgradePlan(address nextImplementation, uint96 delay) external {
-        if (delay == 0) {
-            delay = 1;
+    /// @param delayEpochs Number of epochs from now before the upgrade may occur
+    function announceUpgradePlan(address nextImplementation, uint96 delayEpochs) external {
+        if (delayEpochs == 0) {
+            delayEpochs = 1;
         }
-        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delay);
+        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delayEpochs);
     }
 
     /// @notice Announce a planned upgrade
--- a/service_contracts/src/ServiceProviderRegistry.sol
+++ b/service_contracts/src/ServiceProviderRegistry.sol
@@ -113,6 +113,7 @@ contract ServiceProviderRegistry is
         uint96 afterEpoch;
     }
 
+    // Pending upgrade announcement
     PlannedUpgrade public nextUpgrade;
 
     event UpgradeAnnounced(PlannedUpgrade plannedUpgrade);
@@ -770,12 +771,32 @@ contract ServiceProviderRegistry is
 
     /// @notice Announce a planned upgrade
     /// @dev Can only be called by the contract owner
+    /// @param nextImplementation Address of the new implementation contract
+    /// @param delay Number of epochs from now before the upgrade may occur
+    function announceUpgradePlan(address nextImplementation, uint96 delay) external {
+        if (delay == 0) {
+            delay = 1;
+        }
+        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delay);
+    }
+
+    /// @notice Announce a planned upgrade
+    /// @dev Can only be called by the contract owner
     /// @param plannedUpgrade The planned upgrade details
-    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external onlyOwner {
-        require(plannedUpgrade.nextImplementation.code.length > 3000);
-        require(plannedUpgrade.afterEpoch > block.number);
-        nextUpgrade = plannedUpgrade;
-        emit UpgradeAnnounced(plannedUpgrade);
+    /// @custom:deprecated Use announceUpgradePlan instead
+    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external {
+        uint96 minAfterEpoch = uint96(block.number + 1);
+        _announcePlannedUpgrade(
+            plannedUpgrade.nextImplementation,
+            plannedUpgrade.afterEpoch < minAfterEpoch ? minAfterEpoch : plannedUpgrade.afterEpoch
+        );
+    }
+
+    function _announcePlannedUpgrade(address nextImplementation, uint96 afterEpoch) internal onlyOwner {
+        require(nextImplementation.code.length > 3000);
+        nextUpgrade.nextImplementation = nextImplementation;
+        nextUpgrade.afterEpoch = afterEpoch;
+        emit UpgradeAnnounced(nextUpgrade);
     }
 
     /// @notice Authorizes an upgrade to a new implementation
--- a/service_contracts/test/FilecoinWarmStorageService.t.sol
+++ b/service_contracts/test/FilecoinWarmStorageService.t.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.13;
 
 import {MockFVMTest} from "@fvm-solidity/mocks/MockFVMTest.sol";
+import {stdError} from "forge-std/StdError.sol";
 import {console, Test, Vm} from "forge-std/Test.sol";
 import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
@@ -446,7 +447,7 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         new MyERC1967Proxy(address(serviceImpl4), initDataLongDesc);
     }
 
-    function testUpgrade() public {
+    function testUpgrade(bool useDeprecatedMethod) public {
         FilecoinWarmStorageService firstServiceImpl = new FilecoinWarmStorageService(
             address(mockPDPVerifier),
             address(payments),
@@ -477,6 +478,7 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
 
         bytes memory migrateData = abi.encodeWithSelector(FilecoinWarmStorageService.migrate.selector, viewContract);
 
+        // Upgrade plan should be cleared
         (address nextImplementation, uint96 afterEpoch) = viewContract.nextUpgrade();
         assertEq(nextImplementation, address(0));
         assertEq(afterEpoch, uint96(0));
@@ -495,12 +497,20 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
             4
         );
 
+        // Another successful announcement
+        nextImplementation = address(newServiceImpl);
+        uint96 delay = 2000;
+        afterEpoch = uint96(vm.getBlockNumber()) + delay;
         FilecoinWarmStorageService.PlannedUpgrade memory plan;
-        plan.nextImplementation = address(newServiceImpl);
-        plan.afterEpoch = uint96(vm.getBlockNumber()) + 2000;
-        service.announcePlannedUpgrade(plan);
 
-        (nextImplementation, afterEpoch) = viewContract.nextUpgrade();
+        if (useDeprecatedMethod) {
+            plan.nextImplementation = nextImplementation;
+            plan.afterEpoch = afterEpoch;
+            service.announcePlannedUpgrade(plan);
+        } else {
+            service.announceUpgradePlan(nextImplementation, delay);
+        }
+        (plan.nextImplementation, plan.afterEpoch) = viewContract.nextUpgrade();
         assertEq(nextImplementation, plan.nextImplementation);
         assertEq(afterEpoch, plan.afterEpoch);
 
@@ -509,12 +519,33 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         service.upgradeToAndCall(nextImplementation, migrateData);
         vm.roll(plan.afterEpoch - 1);
         vm.expectRevert();
-        service.upgradeToAndCall(plan.nextImplementation, migrateData);
+        service.upgradeToAndCall(nextImplementation, migrateData);
 
         vm.roll(plan.afterEpoch);
         vm.expectEmit(false, false, false, true, address(service));
-        emit FilecoinWarmStorageService.ContractUpgraded(newServiceImpl.VERSION(), plan.nextImplementation);
-        service.upgradeToAndCall(plan.nextImplementation, migrateData);
+        emit FilecoinWarmStorageService.ContractUpgraded(newServiceImpl.VERSION(), nextImplementation);
+        service.upgradeToAndCall(nextImplementation, migrateData);
+
+        // Plan should be cleared
+        (plan.nextImplementation, plan.afterEpoch) = viewContract.nextUpgrade();
+        assertEq(address(0), plan.nextImplementation);
+        assertEq(0, plan.afterEpoch);
+
+        // Check behavior of minimum delay
+        if (useDeprecatedMethod) {
+            plan.nextImplementation = nextImplementation;
+            plan.afterEpoch = 0;
+            service.announcePlannedUpgrade(plan);
+        } else {
+            // prevent overflow
+            vm.expectRevert(stdError.arithmeticError);
+            service.announceUpgradePlan(nextImplementation, type(uint96).max);
+
+            service.announceUpgradePlan(nextImplementation, 0);
+        }
+        (plan.nextImplementation, plan.afterEpoch) = viewContract.nextUpgrade();
+        assertEq(plan.nextImplementation, nextImplementation);
+        assertEq(plan.afterEpoch, vm.getBlockNumber() + 1);
     }
 
     function _getSingleMetadataKV(string memory key, string memory value)
--- a/service_contracts/test/FilecoinWarmStorageService.t.sol
+++ b/service_contracts/test/FilecoinWarmStorageService.t.sol
@@ -448,7 +448,15 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         new MyERC1967Proxy(address(serviceImpl4), initDataLongDesc);
     }
 
-    function testUpgrade(bool useDeprecatedMethod) public {
+    function testAnnouncePlannedUpgrade() public {
+        _testUpgrade(true);
+    }
+
+    function testAnnounceUpgradePlan() public {
+        _testUpgrade(false);
+    }
+
+    function _testUpgrade(bool useDeprecatedMethod) internal {
         FilecoinWarmStorageService firstServiceImpl = new FilecoinWarmStorageService(
             address(mockPDPVerifier),
             address(payments),
@@ -549,7 +557,15 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         assertEq(plan.afterEpoch, vm.getBlockNumber() + 1);
     }
 
-    function testAnnouncePlannedUpgradeOnlyOwner(bool useDeprecatedMethod) public {
+    function testAnnouncePlannedUpgradeOnlyOwner() public {
+        _testAnnouncePlannedUpgradeOnlyOwner(true);
+    }
+
+    function testAnnounceUpgradePlanOnlyOwner() public {
+        _testAnnouncePlannedUpgradeOnlyOwner(false);
+    }
+
+    function _testAnnouncePlannedUpgradeOnlyOwner(bool useDeprecatedMethod) internal {
         FilecoinWarmStorageService newServiceImpl = new FilecoinWarmStorageService(
             address(mockPDPVerifier),
             address(payments),
--- a/service_contracts/test/FilecoinWarmStorageService.t.sol
+++ b/service_contracts/test/FilecoinWarmStorageService.t.sol
@@ -9,6 +9,7 @@ import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 import {Cids} from "@pdp/Cids.sol";
 import {MyERC1967Proxy} from "@pdp/ERC1967Proxy.sol";
 import {SessionKeyRegistry} from "@session-key-registry/SessionKeyRegistry.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 
 import {CHALLENGES_PER_PROOF, FilecoinWarmStorageService} from "../src/FilecoinWarmStorageService.sol";
 import {FilecoinWarmStorageServiceStateView} from "../src/FilecoinWarmStorageServiceStateView.sol";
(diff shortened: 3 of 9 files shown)


rjan90:
I opened [#548](https://github.com/FilOzone/filecoin-services/pull/548) as a stacked PR targeting this branch to cover the operational follow-through for this change. It:

- Regenerates the published FWSS ABI, fixing the failing ABI check in this PR.
- Makes `announceUpgradePlan(address,uint96)` the default path in the upgrade helper.
- Updates the upgrade checklist to select a relative delay, then read and record the actual on-chain afterEpoch after Safe execution.
- Adds preflight simulation so unsupported selectors, invalid implementations, and overflowing delays are caught before calldata is proposed.
- Preserves an explicit legacy mode for the FWSS v1.3.0 → v1.3.1 rollout, since the currently deployed v1.3.0 contracts do not expose the new selector.

rjan90:
I do wonder if we should update `ServiceProviderRegistry` to also use `announceUpgradePlan(address,uint96) `. Or at least record it as another issue ticket that can be picked up later, so that these contracts are using the same method.

rjan90:
I saw that you closed #548, and agree that we do not need to land the dual-mode script from #548. But there was still a lot of valuable/independent changes that is needed:

- The regenerated FWSS ABI, which is still causing Check ABI Changes to fail on this PR.
- Upgrade-checklist improvements for recording and verifying the actual nextUpgrade() implementation and afterEpoch after the Safe transaction.
- Documentation of the v1.3.1 bootstrap sequence and the subsequent migration of tooling to announceUpgradePlan.

Please review the checklist changes in #548 and fold the applicable parts, along with the ABI update, into this PR?

service_contracts/tools/README.md
```diff
@@ -200,12 +200,13 @@ See [UPGRADE-CHECKLIST.md](./UPGRADE-CHECKLIST.md) for the complete two-step FWS
...
+The FilecoinWarmStorageService and ServiceProviderRegistry contracts use a **two-step upgrade process** for security. The normal FWSS flow is:
 
-1. **Announce**: Call `announcePlannedUpgrade()` with the new implementation address and a future epoch
-2. **Execute**: After the announced epoch, call `upgradeToAndCall()` to complete the upgrade
+1. **Announce**: Call `announceUpgradePlan()` with the new implementation address and a relative delay
+2. **Observe**: Read `nextUpgrade()` after the announcement lands and record its exact `afterEpoch`
+3. **Execute**: After the observed epoch, call `upgradeToAndCall()` to complete the upgrade
 
-This gives stakeholders time to review changes before execution.
+The delay is measured from the block in which the announcement executes, so Safe signing time does not consume the requested notice window. ServiceProviderRegistry continues to use `announcePlannedUpgrade()` with an absolute epoch.
```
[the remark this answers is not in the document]
rjan90: Created followup ticket for this: https://github.com/FilOzone/filecoin-services/issues/550 since the currently deployed registry implementations predate the new selector, the first upgrade that introduces it must still be announced through the legacy interface.

rjan90 (approved):

BigLep:
I will take a look today as well so we can merge this

service_contracts/src/ServiceProviderRegistry.sol
```diff
@@ -768,14 +769,34 @@ contract ServiceProviderRegistry is
...
+    /// @custom:deprecated Use announceUpgradePlan instead
+    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external {
+        uint96 minAfterEpoch = uint96(block.number + 1);
+        _announcePlannedUpgrade(
+            plannedUpgrade.nextImplementation,
+            plannedUpgrade.afterEpoch < minAfterEpoch ? minAfterEpoch : plannedUpgrade.afterEpoch
+        );
+    }
+
+    function _announcePlannedUpgrade(address nextImplementation, uint96 afterEpoch) internal onlyOwner {
```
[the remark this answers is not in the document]
wjmelements: This adds a few hundred bytes to the codesize so I prefer to share it in the internal method. The test coverage will ensure it reverts for non-owners.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -405,11 +406,27 @@ contract FilecoinWarmStorageService is
...
+            plannedUpgrade.afterEpoch < minAfterEpoch ? minAfterEpoch : plannedUpgrade.afterEpoch
+        );
+    }
+
+    function _announcePlannedUpgrade(address nextImplementation, uint96 afterEpoch) internal onlyOwner {
+        require(nextImplementation.code.length > 3000);
+        nextUpgrade.nextImplementation = nextImplementation;
+        nextUpgrade.afterEpoch = afterEpoch;
+        emit UpgradeAnnounced(nextUpgrade);
     }
```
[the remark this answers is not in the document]
wjmelements: This adds a few hundred bytes to the codesize so I prefer to share it in the internal method. The test coverage will ensure it reverts for non-owners.

service_contracts/test/ServiceProviderRegistry.t.sol
```diff
@@ -53,7 +53,7 @@ contract ServiceProviderRegistryTest is MockFVMTest {
         registry.initialize();
     }
 
-    function testAnnouncePlannedUpgrade() public {
+    function testAnnouncePlannedUpgrade(bool useDeprecatedMethod) public {
```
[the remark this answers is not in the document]
wjmelements: e2f390b24b4510ef35b19637614e00abc3bc38ac

service_contracts/test/ServiceProviderRegistry.t.sol
```diff
@@ -115,35 +122,51 @@ contract ServiceProviderRegistryTest is MockFVMTest {
         assertEq(afterEpoch, uint96(0));
     }
 
-    function testAnnouncePlannedUpgradeOnlyOwner() public {
+    function testAnnouncePlannedUpgradeOnlyOwner(bool useDeprecatedMethod) public {
```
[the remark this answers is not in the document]
wjmelements: e2f390b24b4510ef35b19637614e00abc3bc38ac

service_contracts/test/ServiceProviderRegistry.t.sol
```diff
@@ -115,35 +122,51 @@ contract ServiceProviderRegistryTest is MockFVMTest {
...
     }
 
-    function testAnnouncePlannedUpgradeInvalidImplementation() public {
-        ServiceProviderRegistry.PlannedUpgrade memory plan;
-        plan.nextImplementation = address(0x123); // Invalid address with no code
-        plan.afterEpoch = uint96(vm.getBlockNumber()) + 2000;
-
-        vm.expectRevert();
-        registry.announcePlannedUpgrade(plan);
+    function testAnnouncePlannedUpgradeInvalidImplementation(bool useDeprecatedMethod) public {
```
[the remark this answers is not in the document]
wjmelements: e2f390b24b4510ef35b19637614e00abc3bc38ac

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -446,7 +448,7 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
         new MyERC1967Proxy(address(serviceImpl4), initDataLongDesc);
     }
 
-    function testUpgrade() public {
+    function testUpgrade(bool useDeprecatedMethod) public {
```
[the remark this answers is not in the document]
wjmelements: e2f390b24b4510ef35b19637614e00abc3bc38ac

service_contracts/test/FilecoinWarmStorageService.t.sol
```diff
@@ -509,12 +520,56 @@ contract FilecoinWarmStorageServiceTest is MockFVMTest {
...
+            service.announceUpgradePlan(nextImplementation, type(uint96).max);
+
+            service.announceUpgradePlan(nextImplementation, 0);
+        }
+        (plan.nextImplementation, plan.afterEpoch) = viewContract.nextUpgrade();
+        assertEq(plan.nextImplementation, nextImplementation);
+        assertEq(plan.afterEpoch, vm.getBlockNumber() + 1);
+    }
+
+    function testAnnouncePlannedUpgradeOnlyOwner(bool useDeprecatedMethod) public {
```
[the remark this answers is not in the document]
wjmelements: e2f390b24b4510ef35b19637614e00abc3bc38ac

service_contracts/test/ServiceProviderRegistry.t.sol
```diff
@@ -115,35 +122,51 @@ contract ServiceProviderRegistryTest is MockFVMTest {
...
+            registry.announcePlannedUpgrade(plan);
+        } else {
+            vm.expectRevert();
+            registry.announceUpgradePlan(address(0x123), 2000); // Invalid address with no code
+        }
     }
 
-    function testAnnouncePlannedUpgradeInvalidEpoch() public {
+    // A low or past afterEpoch/delay is clamped up to the minimum rather than reverting
+    function testAnnouncePlannedUpgradeMinimumDelay(bool useDeprecatedMethod) public {
```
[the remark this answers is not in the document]
wjmelements: e2f390b24b4510ef35b19637614e00abc3bc38ac

service_contracts/tools/README.md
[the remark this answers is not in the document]
wjmelements: 02c9133f76611548bdf0899c3b508d0c4024f6c6

wjmelements:
@rjan90 fyi I changed the default to use the deprecated method since the new method is not deployed.

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -405,11 +406,27 @@ contract FilecoinWarmStorageService is
         challengeWindowSize = _challengeWindowSize;
     }
 
-    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external onlyOwner {
-        require(plannedUpgrade.nextImplementation.code.length > 3000);
-        require(plannedUpgrade.afterEpoch > block.number);
-        nextUpgrade = plannedUpgrade;
-        emit UpgradeAnnounced(plannedUpgrade);
+    function announceUpgradePlan(address nextImplementation, uint96 delay) external {
```
BigLep: Rename `delay` → `delayEpochs` here and in `ServiceProviderRegistry` for self-documentation. Matches the existing `afterEpoch` / `block.number` convention (ABI regenerates from Solidity).

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -405,11 +406,27 @@ contract FilecoinWarmStorageService is
...
+        if (delay == 0) {
+            delay = 1;
+        }
+        _announcePlannedUpgrade(nextImplementation, uint96(block.number) + delay);
+    }
+
+    /// @custom:deprecated Use announceUpgradePlan instead
+    function announcePlannedUpgrade(PlannedUpgrade calldata plannedUpgrade) external {
```
BigLep, line 417: Do we have a tracking item to remove this deprecated method after v1.3.1 is live on both networks?

service_contracts/src/FilecoinWarmStorageService.sol
BigLep: This is a public external entrypoint but has no comments (unlike `ServiceProviderRegistry.announceUpgradePlan`, which already documents params). Worth adding brief `@notice` / `@param` docs here too?

Also: in #533 you had a shared parent (`UpgradeHardening.sol`). Should we do that here so FWSS and ServiceProviderRegistry share this upgrade-announce logic, or is duplicating the small helpers preferred for this PR?

(Newbie question: in Solidity contracts, sharing via an abstract base seems common for UUPS plumbing. Is "favor composition over inheritance" a norm in contract world, or not really?)

service_contracts/tools/README.md
```diff
@@ -257,12 +258,28 @@ The following scripts support `CALLDATA_ONLY=true`:
...
+export ANNOUNCEMENT_MODE=legacy
+export LEGACY_NOTICE_EPOCHS=2880
+export SAFE_SIGNING_BUFFER_EPOCHS=2880
+CURRENT_EPOCH=$(cast block-number --rpc-url "$ETH_RPC_URL")
+export AFTER_EPOCH=$((CURRENT_EPOCH + SAFE_SIGNING_BUFFER_EPOCHS + LEGACY_NOTICE_EPOCHS))
+unset UPGRADE_DELAY_EPOCHS
+CALLDATA_ONLY=true ./warm-storage-announce-upgrade.sh
+```
+
+Treat this bootstrap-only mode as deprecated after FWSS v1.3.1 is live on both Calibnet and Mainnet. Remove it once rollback to v1.3.0 is no longer supported; until then it remains available only for that rollback path.
```
BigLep, line 281: @rjan90 Let's get this into a tracking issue under the FWSS M4.5 GA follow-up work after the network upgrade — cleanup for both the legacy tooling/docs path and the deprecated Solidity entrypoint once rollback to v1.3.0 is no longer supported.

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -49,21 +49,25 @@ Field ownership for duplicated rollout data:
 
 ### Upgrade Schedule
 
-| Network | AFTER_EPOCH | Status |
-|---|---:|---|
-| Calibnet | `TBD` | Pending |
-| Mainnet | `TBD` | Pending |
+| Network | Announcement mode (v1.3.1 bootstrap only) | Requested delay | Actual `AFTER_EPOCH` | Status |
```
BigLep, line 52:
```suggestion
| Network | Announcement mode (v1.3.1 bootstrap only) | Requested delay epochs | Actual `AFTER_EPOCH` | Status |
```

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -146,15 +151,30 @@ Record validation that proves the planned upgrade works against the full contrac
 | Routine | `2880` epochs (~24h) | 1-2 days |
 | Breaking change | `20160` epochs (~1 week) | 1-2 weeks |
 
-Calibnet can use a shorter window for rehearsal and validation, but use enough time for signers to coordinate.
+Calibnet can use a shorter window for rehearsal and validation, but use enough time for signers to coordinate. Select a positive operational delay; the contract's one-epoch floor is an emergency safety bound, not the routine notice policy.
```
BigLep, line 154:
```suggestion
| Upgrade Type | Minimum Notice | Recommended |
|---|---:|---|
| Calibration | `240` epochs (~2h) | ~1 day |
| Mainnet Routine | `2880` epochs (~24h) | 1-2 days |
| Mainnet Breaking change | `20160` epochs (~1 week) | 1-2 weeks |

Select a positive operational delay; the contract's one-epoch floor is an emergency safety bound, not the routine notice policy.
```

With delay-based announcement, Safe signing time no longer consumes the notice window, so the "enough time for signers to coordinate" wording can go. (Signers still need to coordinate to sign — that just shouldn't dictate the delay.)

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -146,15 +151,30 @@ Record validation that proves the planned upgrade works against the full contrac
...
+
+```bash
+export UPGRADE_DELAY_EPOCHS=2880 # use 240+ for Calibnet rehearsal, 20160 for breaking changes
```
BigLep, line 157:
```suggestion
export UPGRADE_DELAY_EPOCHS=<ENTER_VALUE_FROM_TABLE_ABOVE>
```

I think we should make people deliberate about the value they choose and not just get a value from copy/pasting without thinking.

(Same goes below.)

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -357,15 +377,7 @@ In Safe Transaction Builder, set target to the printed FWSS proxy, value to `0`,
...
-- [ ] Compute Calibnet `AFTER_EPOCH` and update the schedule table
-
-```bash
-export ETH_RPC_URL="https://api.calibration.node.glif.io/rpc/v1"
-export UPGRADE_WAIT_DURATION_EPOCHS=240 # use a longer window if desired
-CURRENT_EPOCH=$(cast block-number --rpc-url "$ETH_RPC_URL")
-AFTER_EPOCH=$((CURRENT_EPOCH + UPGRADE_WAIT_DURATION_EPOCHS))
-echo "$CURRENT_EPOCH -> $AFTER_EPOCH"
-```
+- [ ] Set the Calibnet requested delay and update the schedule table. **v1.3.1 bootstrap only:** record the announcement mode as `legacy`; upgrades from v1.3.1 onward always use `delay`.
```
BigLep, line 380:
```suggestion
- [ ] Set the requested delay in the update table.. **v1.3.1 bootstrap only:** record the announcement mode as `legacy`; upgrades from v1.3.1 onward always use `delay`.
```

I think this step is just about updating the table.

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -473,17 +544,8 @@ The unique `smoke_run` metadata is required so this validates new Data Set creat
...
-- [ ] Compute Mainnet `AFTER_EPOCH` and update the schedule table
-
-```bash
-export ETH_RPC_URL="https://api.node.glif.io/rpc/v1"
-export UPGRADE_WAIT_DURATION_EPOCHS=2880 # use 20160 for breaking changes
-CURRENT_EPOCH=$(cast block-number --rpc-url "$ETH_RPC_URL")
-AFTER_EPOCH=$((CURRENT_EPOCH + UPGRADE_WAIT_DURATION_EPOCHS))
-echo "$CURRENT_EPOCH -> $AFTER_EPOCH"
-```
+- [ ] Set the Mainnet requested delay and update the schedule table. **v1.3.1 bootstrap only:** record the announcement mode as `legacy`; upgrades from v1.3.1 onward always use `delay`.
```
BigLep, line 549: Same as Calibnet: the steps below cover env/calldata generation, but please keep an explicit checkbox to update the schedule table with the requested delay and observed `AFTER_EPOCH`.

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -492,17 +554,76 @@ cd service_contracts/tools
...
+For the v1.3.0 -> v1.3.1 bootstrap rollout only, use this configuration instead:
+
+```bash
+export ANNOUNCEMENT_MODE=legacy
+export LEGACY_NOTICE_EPOCHS=2880
+export SAFE_SIGNING_BUFFER_EPOCHS=2880
+CURRENT_EPOCH=$(cast block-number --rpc-url "$ETH_RPC_URL")
+export AFTER_EPOCH=$((CURRENT_EPOCH + SAFE_SIGNING_BUFFER_EPOCHS + LEGACY_NOTICE_EPOCHS))
+unset UPGRADE_DELAY_EPOCHS
+```
```
BigLep, line 577: There is duplication of this logic for calibration and mainnet. A couple of ideas:
1. is this functionality that can just be put into the warm-storage-announce-upgrade script?
2. or do we put the snippets at the bottom of the checklist and reference them?

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -586,6 +707,7 @@ The unique `smoke_run` metadata is required so this validates new Data Set creat
 
 ### Phase 5: Promote Release and Close Out
 - [ ] Confirm live Calibnet and Mainnet FWSS implementation slots match the new implementation addresses
+- [ ] After FWSS v1.3.1 is live on Calibnet and Mainnet, treat `ANNOUNCEMENT_MODE=legacy` as deprecated and decide whether rollback to v1.3.0 is still supported. Once that rollback path is retired, open and merge a follow-up PR that removes the legacy mode, its `AFTER_EPOCH` handling, the temporary announcement-mode schedule column and bootstrap clauses, the README bootstrap example, and the Temporary Bootstrap Compatibility instructions; record the cleanup PR link. If v1.3.0 rollback remains supported, retain legacy mode or document the exact v1.3.1-tagged helper that operators must use.
```
BigLep, line 711: This reads like the in-checklist backlog reminder — fine if that's intentional so we don't lose it, but maybe track the post-rollout cleanup (legacy tooling/docs + deprecated Solidity method) in a dedicated issue linked from https://github.com/FilOzone/filecoin-services/issues/549?

service_contracts/tools/warm-storage-announce-upgrade.sh
```diff
@@ -42,28 +56,103 @@ if [ -z "$NEW_FWSS_IMPLEMENTATION_ADDRESS" ]; then
...
-  echo "AFTER_EPOCH is not set"
+if [ -z "$FWSS_PROXY_ADDRESS" ]; then
+  echo "Error: FWSS_PROXY_ADDRESS is not set"
   exit 1
 fi
 
-CURRENT_EPOCH=$(cast block-number 2>/dev/null)
+is_non_negative_integer() {
+  [[ "$1" =~ ^(0|[1-9][0-9]*)$ ]]
+}
```
BigLep, line 66: Eek! We keep piling on more ugly bash. Really need to move out from it... I know we have the backlog item so I won't decry more here.

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -357,15 +377,7 @@ In Safe Transaction Builder, set target to the printed FWSS proxy, value to `0`,
...
-```bash
-export ETH_RPC_URL="https://api.calibration.node.glif.io/rpc/v1"
-export UPGRADE_WAIT_DURATION_EPOCHS=240 # use a longer window if desired
-CURRENT_EPOCH=$(cast block-number --rpc-url "$ETH_RPC_URL")
-AFTER_EPOCH=$((CURRENT_EPOCH + UPGRADE_WAIT_DURATION_EPOCHS))
-echo "$CURRENT_EPOCH -> $AFTER_EPOCH"
-```
+- [ ] Set the Calibnet requested delay and update the schedule table. **v1.3.1 bootstrap only:** record the announcement mode as `legacy`; upgrades from v1.3.1 onward always use `delay`.
 
 - [ ] Generate announce calldata and submit/sign/execute in Safe UI:
```
BigLep, line 382:
```suggestion
- [ ] Generate announce calldata
```

It doesn't seem like we're submitting/signing/executing in the Safe UI here.

service_contracts/tools/UPGRADE-CHECKLIST.md
```diff
@@ -374,17 +386,76 @@ cd service_contracts/tools
...
+unset UPGRADE_DELAY_EPOCHS
+```
+
+Generate the transaction after selecting exactly one configuration above:
+
+```bash
 CALLDATA_ONLY=true ./warm-storage-announce-upgrade.sh
 ```
 
 - [ ] In Safe Transaction Builder, set target to the printed FWSS proxy, value to `0`, and data to the printed calldata
```
BigLep, line 416: dont need to expand scope, but doesn't the Safe UI have a CLI for doing this?

BigLep (changes requested):
Some comments on the code and the docs/process @wjmelements and @rjan90

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: > in Solidity contracts, sharing via an abstract base seems common for UUPS plumbing. Is "favor composition over inheritance" a norm in contract world, or not really?

Solidity has multiple inheritance, for better or for worse, so inheritance is a normal way to do composition. The other way is libraries. Libraries have the advantage that they can be `external`, but there isn't a big difference between `internal` functions and inherited functions. Libraries can only access state via `storage` pointers, but such pointers can also allow superclasses to read and write the state of the subclasses. Libraries can't declare their own state, but they often do anyway using patterns such as ERC-7201. One thing they actually can't do is `override` or declare `virtual` functions. Libraries also can't have `immutable` constants. However, the best reason to prefer inheritance for this is that **libraries can't declare interface methods**. Because we want a common interface, that logic belongs in a superclass.

> Should we do that here so FWSS and ServiceProviderRegistry share this upgrade-announce logic, or is duplicating the small helpers preferred for this PR?

I created the UpgradeHardening for the other design because I wanted that pattern to get widely adopted in order to improve the security of future contracts. I do not feel that way about this pattern, so I leave the triplication as-is.

BigLep: > > Should we do that here so FWSS and ServiceProviderRegistry share this upgrade-announce logic, or is duplicating the small helpers preferred for this PR?
> 
> I created the UpgradeHardening for the other design because I wanted that pattern to get widely adopted in order to improve the security of future contracts. I do not feel that way about this pattern, so I leave the triplication as-is.

I understand that we don't have the base contract-enforced delay pattern yet that we ideally want (that comes with https://github.com/FilOzone/filecoin-services/issues/541), but we could at least move in the direction of having shared logic so:

1. less duplciation now
2. when we do add it, we're adding it one place. 

I believe the changes we're making here are still good even for our future desired end-state of contract enforced (i.e., even in future we're going to want to be able to set `delayEpochs` for our upgradeable contracts).

service_contracts/tools/UPGRADE-CHECKLIST.md
rjan90: Fixed in: https://github.com/FilOzone/filecoin-services/commit/eee19d74285de35680cca79697d6706040830057

service_contracts/tools/UPGRADE-CHECKLIST.md
rjan90: Fixed in: https://github.com/FilOzone/filecoin-services/commit/97e52425edcded8a2eda6c26e84de8a844422fa0

service_contracts/tools/UPGRADE-CHECKLIST.md
rjan90: Fixed in: https://github.com/FilOzone/filecoin-services/commit/97e52425edcded8a2eda6c26e84de8a844422fa0

service_contracts/src/FilecoinWarmStorageService.sol
rjan90: Added tracking item here: https://github.com/FilOzone/filecoin-services/issues/554

service_contracts/tools/UPGRADE-CHECKLIST.md
rjan90: Fixed in: https://github.com/FilOzone/filecoin-services/commit/6f9cb05862d40144394ae9e073bb255b4ef80df4

service_contracts/tools/UPGRADE-CHECKLIST.md
rjan90: In https://github.com/FilOzone/filecoin-services/commit/c6ec2d611cf167966effca15c471b5eb99ea5b5d I put snippets towards the bottom of the checklist and referenced that instead of having the duplication in the checklist

BigLep:
@wjmelements : it looks like the open items are:
1. Rename delay → delayEpochs
2. Decision on copy/paste duplication or start the work of factoring our our common contract upgrade logic (my vote).

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: > when we do add it

I am skeptical of the priority of that task.

> the changes we're making here are still good

It is a tradeoff.

> move in the direction of having shared logic

Because the security provided by this proposal mechanism is tenuous, it should not be reused yet, so I want to delay that refactor.

service_contracts/tools/UPGRADE-CHECKLIST.md
wjmelements: I think we prefer the transaction builder over the CLI.

service_contracts/src/FilecoinWarmStorageService.sol
wjmelements: [1b7be66](https://github.com/FilOzone/filecoin-services/pull/547/commits/1b7be66fbe5a075da73faa12d4612dd425f7861e)

BigLep (approved):
Lets merge this puppy (and I guess add the same pattern to pdp).

PR #548, opened 2026-07-10
fix: align FWSS upgrade tooling with delay plans

rjan90:
## Summary

- make `announceUpgradePlan(address,uint96)` the default FWSS operator flow
- keep an explicit bootstrap-only `ANNOUNCEMENT_MODE=legacy` path for the FWSS v1.3.0 -> v1.3.1 rollout
- simulate announcements from the proxy owner before printing or sending calldata
- record and verify the actual on-chain `afterEpoch` in the upgrade checklist and release run log
- regenerate the published FWSS ABI

## Why

This is stacked on #547. Relative delays remove the need to predict an absolute epoch before Safe signing, but the currently deployed FWSS v1.3.0 contracts do not expose the new selector. Their upgrade to v1.3.1 must therefore use the legacy interface.

Treat legacy mode as deprecated after v1.3.1 is live on Calibnet and Mainnet. Remove it once rollback to v1.3.0 is no longer supported; if that rollback remains supported, retain the mode or document the exact v1.3.1-tagged helper operators must use.

### DIFF
--- a/service_contracts/abi/FilecoinWarmStorageService.abi.json
+++ b/service_contracts/abi/FilecoinWarmStorageService.abi.json
@@ -106,6 +106,24 @@
   },
   {
     "type": "function",
+    "name": "announceUpgradePlan",
+    "inputs": [
+      {
+        "name": "nextImplementation",
+        "type": "address",
+        "internalType": "address"
+      },
+      {
+        "name": "delay",
+        "type": "uint96",
+        "internalType": "uint96"
+      }
+    ],
+    "outputs": [],
+    "stateMutability": "nonpayable"
+  },
+  {
+    "type": "function",
     "name": "configureProvingPeriod",
     "inputs": [
       {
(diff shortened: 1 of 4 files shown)


wjmelements:
We can't change the scripts to use the non existent method until executing the upgrade that adds it. This is also why I deprecate but do not remove the method: so we can still use the scripts.

rjan90:
Understood on keeping the existing script for the v1.3.0 → v1.3.1 rollout.

There are still useful checklist changes in #548 that do not depend on switching the script, particularly reading back and verifying nextUpgrade() after the Safe transaction, recording the actual afterEpoch in the schedule/run log, and documenting the post-v1.3.1 tooling transition.

Could you review those portions and either fold the applicable changes into #547 or keep them as a smaller follow-up?

wjmelements (approved):

PR #551, opened 2026-07-13
feat: introduce IFilecoinServiceMetadata and apply it in FWSS

### LINKED ISSUE #535: Reopen #196: store service name + description on-chain as state
## Description

Reverse [#196](https://github.com/FilOzone/filecoin-services/pull/196) (re-applied in [#199]((https://github.com/FilOzone/filecoin-services/pull/199)). Put `name` and `description` back in contract state so an operator can be identified with a single `eth_call`. Keep the `FilecoinServiceDeployed` event for indexing — just not as the only source of truth.

I approved #196 and think it was the wrong call. Reasons below.

## Why

- **Identity should be readable on-chain.** Name/description are a static per-contract attribute. Emit-only means every consumer must trust an indexer to index `FilecoinServiceDeployed` — so it's currently unresolvable easily by anyone.
- **Hard-coding and indexing don't scale to multi-operator services.** FWSS is one operator, so a hand-maintained map looks fine for the pay explorer — that's the easy case. [FCSS deploys a separate operator contract per client deal](https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1782122860946539) , so operator addresses are unbounded. Whether FCSS's approach wrt its operator contracts is optimal, its calling out a gap for our explorer:  static `known-addresses.ts` map is hopeless when FOC scales even an indexer must catch every deploy. Self-describing contracts invert it: the address tells you what it is when you ask it. That's the only primitive that scales.
- **Getters are the convention for identity** (ERC-20 `name()`/`symbol()`, ERC-721, ENS). Emit-only is for data you never read back.
- **Cost is negligible** — set once at init, ≤256 bytes each. We traded composability for trivial gas.

## Acceptance criteria

1. **`name` and `description` stored in contract state** on `FilecoinWarmStorageService`, set in `initialize()` (existing validation kept). Public state vars auto-expose a view getter, so they're resolvable via `eth_call` — no separate getter or setter needed. If a value ever needs to change, do it via contract upgrade rather than a dedicated setter. Keep emitting `FilecoinServiceDeployed`.
2. **Quick FRC draft** standardizing this as a metadata convention for FOC service contracts (storage field names + types + constraints), so any service operator — FWSS, FCSS, future ones — implements it the same way and shows up correctly everywhere with no dashboard changes.

## Open question

- wether to use https://eips.ethereum.org/EIPS/eip-165 
- Storage-layout impact of the restored slots on the existing upgradeable proxy.

rvagg:
Closes: https://github.com/FilOzone/filecoin-services/issues/535

Need's @jennijuju's sign-off. We talked over the various permutations about this and she was going to decide what she wants to consume for the console.

My words of caution about this:

* Untrusted from any contract you don't know, always escape all output and even consider hiding because you never know what it's going to contain. Links shouldn't be clickable etc.
* For contracts you do trust, it's not really critical you have this because you presumably know what that contract is, but this would give those contracts the ability to adjust their own branding .. which implies you _trust_ them enough to defer these strings to them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@ Storage providers submit on-chain transactions on behalf of clients (piece addit
 
 The complete on-chain price catalogue is exposed via `FilecoinWarmStorageServiceStateView.getPriceList()`. It returns a single nested `PriceList` struct covering token, streaming rates, one-time fees, and lockup amounts/periods. See [SPEC.md](SPEC.md) for details on rate calculation, operation fees, pricing updates, and top-up/renewal behavior.
 
+## Service metadata
+
+FilecoinWarmStorageService implements `IFilecoinServiceMetadata`, exposing a `name()` and `description()` for `eth_call`. Other FOC services are encouraged to use this interface so explorers and clients can identify service operator contracts without hard-coded address maps. Consumers should note that this output should be treated as untrusted data and should only ever be presented to users in carefully escaped form.
+
 ## 🚀 Quick Start
 
 ### Prerequisites
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The complete on-chain price catalogue is exposed via `FilecoinWarmStorageService
 
 ## Service metadata
 
-FilecoinWarmStorageService implements `IFilecoinServiceMetadata`, exposing a `name()` and `description()` for `eth_call`. Other FOC services are encouraged to use this interface so explorers and clients can identify service operator contracts without hard-coded address maps. Consumers should note that this output should be treated as untrusted data and should only ever be presented to users in carefully escaped form.
+FilecoinWarmStorageService implements `IFilecoinServiceMetadata`, exposing `name()`, `description()`, and `homepage()` for `eth_call`. Other FOC services are encouraged to use this interface so explorers and clients can identify service operator contracts without hard-coded address maps. `description()` is capped at 256 bytes and should be treated as untrusted display-only text. `homepage()` is an optional URL, capped at 256 bytes, and returns an empty string when not provided. Consumers should validate metadata from unverified contracts and carefully escape all displayed values.
 
 ## 🚀 Quick Start
 
@@ -148,4 +148,3 @@ See [service_contracts/CONTRIBUTING.md](./service_contracts/CONTRIBUTING.md) for
 
 ## 📄 License
 Dual-licensed under [MIT](https://github.com/FilOzone/filecoin-services/blob/main/LICENSE.md) + [Apache 2.0](https://github.com/FilOzone/filecoin-services/blob/main/LICENSE.md)
-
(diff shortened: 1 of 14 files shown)


rjan90:
https://github.com/FilOzone/filecoin-services/issues/535#issuecomment-5088406960

rjan90:
Chatted with Jennifer about the intended scope of the v1 Filecoin service metadata standard. The ask was that the initial version expose three fields:

- `name()` — a short service label
- `description()` — a concise description, capped at approximately 256 bytes and treated as untrusted display-only text
- `homepage()` — an optional, size-capped URL for technical documentation, specifications, or the service repository

I’ve created [PR #559](https://github.com/FilOzone/filecoin-services/pull/559), stacked directly on this branch, which adds `homepage()`, updates the implementation and tests, and regenerates the affected ABIs.

wjmelements:
There's still no reason to do this. At worst it provides an interface for misleading the public. At best it add information everyone already knows to a place nobody will see unless they already knew that information. The IERC20 comparison is wrong; services aren't composable like tokens because there is no common interface for interacting with them. This metadata doesn't help anyone interact with it on-chain, so it doesn't belong on-chain.

Can we please block this until we get a solc with my dispatcher fix? https://github.com/argotorg/solidity/pull/16845 That would mitigate the gas impact on the other functions (though we'd still have to pay for the codesize increase).

wjmelements:
What you really want is a curated registry with this information. That would provide the same information in a discoverable way but it wouldn't be self-identification, and would avoid all of the problems.

jennijuju:
> There's still no reason to do this. At worst it provides an interface for misleading the public. At best it add information everyone already knows to a place nobody will see unless they already knew that information. The IERC20 comparison is wrong; services aren't composable like tokens because there is no common interface for interacting with them. This metadata doesn't help anyone interact with it on-chain, so it doesn't belong on-chain.
> 
> Can we please block this until we get a solc with my dispatcher fix? [argotorg/solidity#16845](https://github.com/argotorg/solidity/pull/16845) That would mitigate the gas impact on the other functions (though we'd still have to pay for the codesize increase).

we need this info to make pay console get human readable info to the user. "Information everyone already knows" holds only for the single-operator FWSS case. Even for tokens, the comparison is to the metadata trio, not transfer semantics. name()/symbol() are likewise never consumed on-chain and likewise untrusted (every  scam token calls itself USDC) — they exist so any wallet or explorer can render any token without a bespoke list. That is exactly what we need here,  for user facing tooling, not for contract developers and  composability was  never what those methods were for.

rjan90:
Rebased this PR onto the latest `main` and resolved the conflicts; all checks are green, so it should be ready for another review.

That said, I see there is still some discussion around these changes. It would be good to reach a decision soon, as this is currently gating the M4.5 contract upgrade: either we land it before the upgrade proceeds or agree to defer it.

jennijuju:
@rjan90 Will and I have talked through  the concern and how we will use this in our tooling, and we are okay to move forward now!

rjan90 (approved):

PR #552, opened 2026-07-13
Special case pre-activation epochs for settlement & preserve activation epoch across reactivation

rvagg:
This follows on from https://github.com/FilOzone/filecoin-services/pull/545 and arose while reviewing the complexity Curio currently needs to handle pre-activation settlement in https://github.com/filecoin-project/curio/pull/1346. Most of that complexity should become unnecessary once this is deployed.

There is also a latent reactivation bug whose payment consequences are made worse by the first fix, so the second commit addresses that as part of the same activation-epoch model.

1. The first commit deals with pre-activation settlement: Data set activation (the first `nextProvingPeriod()` call) typically follows rate change activity, notably piece additions (combined create+add is the norm). So likely we've already adjusted the rate on the PDP rail. Filecoin Pay can therefore present FWSS with rate segments that end before activation, which FWSS previously could not validate cleanly. A client may also terminate before the SP calls `nextProvingPeriod()`, further exposing this case. From a billing perspective, streaming payment should not begin until the SP signals that proving has started. Ranges ending at or before the activation epoch shoudl settle with zero payment. Ranges _crossing_ activation exclude the pre-activation portion and validate the remainder normally. So, we clearly say that streaming payment eligibility begins at `activationEpoch + 1` and remains proof-dependent. This generalises the fix from https://github.com/FilOzone/filecoin-services/pull/545.
2. The second commit deals with the latent reactivation bug: When all pieces are removed, the data set becomes inactive by setting `provingDeadlines[dataSetId] = 0`, but its proof bitmap and payment history remain. Previously, the first `nextProvingPeriod()` after pieces were re-added overwrote `provingActivationEpoch` with the current block, rebasing the proving timeline and restarting period numbering from zero. This can make legitimate historical settlement fail with `InvalidEpochRange`. With the first commit, those historical epochs could instead be misclassified as pre-activation and settled with zero payment. Restarting period numbering could also allow aliasing of retained proof bits (!), making a reactivated period appear proven before the newly added pieces have been proved. The fix treats `provingActivationEpoch` as the immutable lifetime origin and uses `provingDeadlines == 0` only to represent temporary proving suspension. "Reactivation" must resume on a fresh proving period aligned with the original timeline. `nextPDPChallengeWindowStart()` now supplies that aligned future window, which Curio should use for reactivation once this contract version is deployed.

Clear as mud?

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -96,7 +96,7 @@ Only the contract owner can update pricing, by upgrading the contract.
 
 Rate recalculation timing differs for additions and deletions due to proving semantics:
 
-- **Adding pieces**: The rate updates immediately when `piecesAdded()` is called. The client begins paying for new pieces right away, even though those pieces won't be included in proof challenges until the next proving period. This fail-fast behavior protects providers: if the client lacks sufficient funds for the new lockup, the transaction fails before the provider commits resources.
+- **Adding pieces**: The rate and corresponding lockup requirement update immediately when `piecesAdded()` is called, even before proving activation. This fail-fast behavior protects providers by rejecting additions the client cannot fund before the provider commits resources. Streaming storage payment becomes eligible only after the first `nextProvingPeriod()` activates proving; pre-activation epochs settle with zero payment. One-time lifecycle fees remain payable before activation.
 
 - **Removing pieces**: Deletions are scheduled and take effect at the next proving boundary (`nextProvingPeriod()`). The client continues paying the existing rate until the removal is finalized. This deferral is required because proofs may challenge any portion of the current data set during the proving period—the provider must continue storing and proving all existing data until the period ends.
 
@@ -104,11 +104,11 @@ Rate recalculation timing differs for additions and deletions due to proving sem
 
 During each proving period, proofs are generated over a fixed data set. The prover must maintain the complete data set because challenges can target any leaf:
 
-- **Additions expand the proof space** but don't affect existing challenges. New pieces simply won't be challenged until the next period. Payment starts immediately because storage resources are committed.
+- **Additions expand the proof space** but don't affect existing challenges. Before activation they update the rate and lockup without earning streaming payment. After activation the new rate becomes payable immediately, even though the added pieces won't be challenged until the next period.
 
 - **Deletions would shrink the proof space** mid-period, potentially invalidating challenges. The data must remain intact until `nextProvingPeriod()` finalizes the removal. Only then does the rate decrease.
 
-This ensures proof integrity while providing fair payment semantics: you pay when you add, and continue paying for deletions until the proving period boundary.
+This ensures proof integrity while providing fair payment semantics: additions reserve funding immediately and become payable once proving is active, while deletions remain payable until the proving period boundary.
 
 ### Rate Changes After Termination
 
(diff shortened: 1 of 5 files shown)


service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -1489,11 +1512,11 @@ contract FilecoinWarmStorageService is
...
-        // discharge lockup. Hit by never-activated data sets and by the finalize leg of
-        // abandonment (which wipes activationEpoch between settles to unblock the open period)
+        // No active proving period covers epochs through the activation boundary. Advance
+        // settlement with zero payment so FilecoinPay can discharge pre-activation rate
+        // segments, including segments recorded before the first nextProvingPeriod call.
         uint256 activationEpoch = provingActivationEpoch[dataSetId];
-        if (activationEpoch == 0) {
+        if (activationEpoch == 0 || toEpoch <= activationEpoch) {
             return ValidationResult({modifiedAmount: 0, settleUpto: toEpoch, note: "No proving activity"});
         }
```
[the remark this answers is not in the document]
rvagg: `FilecoinPayV1.settleRailInternal()` already enforces `require(untilEpoch <= block.number, ...);` so this is covered

service_contracts/src/FilecoinWarmStorageService.sol
```diff
@@ -920,19 +920,42 @@ contract FilecoinWarmStorageService is
...
+                // guarantees a future challenge epoch; derive its canonical deadline without
+                // rebasing payment history or proven-period IDs.
+                require(
+                    challengeEpoch > activationEpoch,
+                    Errors.InvalidChallengeEpoch(
+                        dataSetId, activationEpoch + 1, activationEpoch + maxProvingPeriod, challengeEpoch
+                    )
+                );
+                uint256 period = _provingPeriodForEpoch(activationEpoch, challengeEpoch, maxProvingPeriod);
+                firstDeadline = _calcPeriodDeadline(activationEpoch, period);
```
wjmelements: Because `firstDeadline` is calculated using the period of the `challengeEpoch`, the `challengeEpoch > firstDeadline` check below won't prevent an arbitrarily-high `challengeEpoch`.

wjmelements: https://github.com/FilOzone/filecoin-services/pull/553

wjmelements (approved):

rvagg:
Followed up with Curio specific lens @ https://github.com/filecoin-project/curio/issues/1356

SPEC.md
```diff
@@ -148,6 +148,26 @@ Deposits extend the duration without changing the rate (unless adding pieces tri
...
+
+Because the view call and transaction use their respective current epochs, the returned window can become stale before inclusion if the earliest valid deadline advances. A caller receiving `InvalidChallengeEpoch` must query `nextPDPChallengeWindowStart()` again and retry. The challenge must also satisfy PDPVerifier's normal minimum and maximum delay rules; the current mainnet and calibnet parameters provide ample headroom.
+
+Before activation, `piecesAdded()` still updates the Filecoin Pay rate and lockup requirement. This enforces funding before the provider commits storage, but does not make pre-activation epochs payable. When Filecoin Pay later presents a rate segment ending at or before `A`, `validatePayment()` advances `settleUpto` through the segment with zero payment. Segments crossing `A` exclude their pre-activation epochs from payment. This also lets terminated, never-activated data sets release their streaming lockup cleanly.
+
+One-time lifecycle fees are independent of proving activation and remain payable when their operations occur.
```
ZenGround0, line 169: Thanks for the write up

PR #553, opened 2026-07-13
fix: bound reactivation challenge epoch

wjmelements:
Reviewer @rvagg
Fixes [unbounded `challengeEpoch`](https://github.com/FilOzone/filecoin-services/pull/552/changes#r3573821198)
#### Changes
* calculate `firstDeadline` from `minimumDeadline` instead of `challengeEpoch`


rvagg (approved):
👍 
it's not _quite_ arbitrary, PDPVerifier does constrain it to `MAX_FINALITY` already, but that's still 30 days

PR #555, opened 2026-07-27
chore(ci): upgrade forge to 1.7.1 and fix lint CI

wjmelements:
Updates forge from 1.3.5 to 1.7.1.

This changes the behavior of `forge fmt`, and adds new `forge lint` rules. Some of the new lint rules failed. I fixed one of them (`unwrapped-modifier-logic`) and added the others to the ignore list. We can fix `custom-errors` separately but the other disabled rules are overzealous.


service_contracts/src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
```diff
@@ -231,12 +229,12 @@ library FilecoinWarmStorageServiceStateInternalLibrary {
         returns (bool)
     {
         return uint256(
-            service.extsload(
```
rvagg, line 234: I noticed these formatting diffs when I was accidentally running a newer forge locally for a while; and I frankly hate this new formatting, what's going on here with the indenting choices?? I guess we can live with it but it's going to be annoying to look at I reckon.

service_contracts/foundry.toml
```diff
@@ -27,11 +27,17 @@ remappings = [
...
+    "block-timestamp",
+    "custom-errors",
     "incorrect-shift",
     "mixed-case-function",
     "mixed-case-variable",
+    "multi-contract-file",
     "pascal-case-struct",
     "screaming-snake-case-immutable",
+    "unsafe-cheatcode",
+    "unsafe-typecast",
```
rvagg, line 42: same comment as in pdp, of all of t he excludes, this seems like a good one to not exclude and just fix

wjmelements: as explained in dm hours ago, not blocking this PR for any of these

github.com/foundry-rs/foundry/issues/13255

rvagg (approved):
approving pending opening an issue to remove unsafe-typecast as a separate batch of work for someone

service_contracts/src/lib/FilecoinWarmStorageServiceStateInternalLibrary.sol
wjmelements: Nothing we can do about it except downgrade forge.

wjmelements:
* https://github.com/FilOzone/filecoin-services/issues/556
* https://github.com/FilOzone/filecoin-services/issues/557

PR #565, opened 2026-08-03
fix: use metadata-aware deployment for contract upgrades

rjan90:
## Summary

- delete the standalone ServiceProviderRegistry deploy script and route the Deploy Contract workflow through `tools/warm-storage-deploy-all.sh`
- make reviewed deployment metadata the rollout scope: SPR is included, while PDPVerifier, FilecoinPay, StateView, and all existing proxies are pinned/preserved
- add an upgrade mode that fails before deployment if live code, proxy implementation slots, FWSS dependency getters, StateView, or initializer counters do not match the recorded deployment
- make the dry-run produce the metadata-selected inventory and make workflow failures propagate through `tee`
- verify updated deployment bytecode metadata after a live stack run and update the reusable upgrade checklist to use the single stack path

## Why this unblocks the rollout

The standalone SPR path bypassed `needs_deployment`, so it could drift from the reviewed release scope and required operators to choose components manually. Consolidating on the metadata-aware stack script makes the frozen release metadata authoritative and prevents an SPR fix from accidentally deploying PDP, replacing a proxy, or switching the unchanged StateView. Once this lands, we can re-freeze the v1.3.1 SHA and start the Calibnet deployment plan.

## Expected v1.3.1 inventory

The read-only plan returns the same inventory on Calibnet and Mainnet:

| Decision | Components |
|---|---|
| Deploy | ServiceProviderRegistry implementation, Rails, FilecoinWarmStorageService implementation |
| Preserve | PDPVerifier, FilecoinPay, SignatureVerificationLib, SessionKeyRegistry, StateView, EndorsementSet, and every proxy |

Live counters were read and validated as:

- SPR: `2 -> 3` on both networks
- FWSS: `8 -> 9` on Calibnet and `4 -> 5` on Mainnet

The stack deployment only creates candidate contracts. Existing proxy upgrades and any future StateView switch remain separate Safe-owned actions.

## Validation

- Calibnet metadata plan against the live RPC: passed with the expected inventory
- Mainnet metadata plan against the live RPC: passed with the expected inventory
- injected counter-read failure: exited before `forge` was called
- `forge test --offline --match-contract 'ServiceProviderRegistry.*Test'`: 78 passed, 0 failed
- Bash syntax, workflow YAML parse, JSON parse, deployment address checksums, and `git diff --check`: passed

## After merge

1. Fast-forward/recreate `release-v1.3.1` from the resulting `main` commit and update the frozen candidate SHA in #561.
2. Update the already-rendered #561 SPR exception: replace the deleted dedicated workflow with the `Warm Storage stack` dry-run/live run. Keep the separate SPR announce/execute gates.
3. Run and approve the Calibnet dry-run before any live deployment.


wjmelements:
I think this script should be deleted. It doesn't call `needs_deployment`; manually identifying which things need to be upgraded is dangerous. The upgrade flow should just call `tools/warm-storage-deploy-all.sh`.

rjan90:
Agreed — this PR now deletes the standalone SPR deploy script and routes the rollout through `warm-storage-deploy-all.sh`.

The reason to land this before continuing v1.3.1 is that the standalone path bypassed `needs_deployment` and made operators select upgrade components manually. The stack path makes the reviewed deployment metadata the source of truth, includes SPR, and explicitly preserves PDP, StateView, and all proxies. Once merged, we can re-freeze the release SHA and start the Calibnet plan instead of proceeding through a stale/manual deployment path.

service_contracts/tools/warm-storage-deploy-all.sh
```diff
@@ -227,10 +253,157 @@ deploy_endorsements_if_needed() {
...
+        return 1
+    fi
+    printf '%s' "$counter"
+}
+
 # ========================================
 # Validation
 # ========================================
 
+if [ "$DEPLOYMENT_MODE" = "upgrade" ]; then
```
wjmelements, line 334: this is a good idea

service_contracts/tools/warm-storage-deploy-all.sh
```diff
@@ -299,11 +472,6 @@ deploy_implementation_if_needed \
     "FilecoinPayV1"
 
 # Step 2: Deploy or use existing PDPVerifier implementation
-if [ -n "$PDP_VERIFIER_PROXY_ADDRESS" ]; then
-    PDP_INIT_COUNTER=$($SCRIPT_DIR/get-initialized-counter.sh $PDP_VERIFIER_PROXY_ADDRESS)
-else
-    PDP_INIT_COUNTER=0
-fi
```
wjmelements, line 306: good idea to consolidate these

wjmelements (approved):

Kubuxu (approved):

